PR #45, opened 2023-02-17
Improve logging and reliability

ali-bahjati:
- Add a key to differenciate primary vs secondary network logging
- Set timeout for the network.


src/bin/agent.rs
```diff
@@ -32,7 +32,7 @@ async fn main() {
     let logger = slog::Logger::root(
         slog_async::Async::default(
             LogBuilder::new(
-                slog_term::CompactFormat::new(slog_term::TermDecorator::new().stdout().build())
+                slog_term::FullFormat::new(slog_term::TermDecorator::new().stdout().build())
```
tompntn, line 35: Just curious - why the change?

tompntn (approved):
🙇

src/bin/agent.rs
ali-bahjati: I don't know why it was not showing the key/value passed to it via the parent logger.

PR #46, opened 2023-02-21
Bump version to 1.0.0

ali-bahjati:
As we are starting to roll this out for production the version should become 1.0.0 to indicate that it is a production-ready code.


guibescos (approved):
:ship:

PR #48, opened 2023-03-15
Add a dashboard in a new metrics.rs module

drozdziak1:
# Overview
This change introduces a simple HTML table for quick reference about individual symbol status. The table is served using a new HTTP endpoint, under "/dashboard". 

# Screenshot
[click here](https://imgur.com/a/kRtTGUx)

# Changes to existing code
* The HTTP endpoint is spawned in `agent.rs` together with other async jobs. It reuses the logger and some of the channel tx's.
* Macro recursion limit is raised to 256 (the new type-checked HTML crate is macro-heavy)
* New dependencies
* * typed-html - ergonomic HTML manipulation

# New code
* `metrics.rs` - a new metrics- and dashboard oriented module. I plan to also add Prometheus metrics there and I'll probably exile dashboard generation logic to something like `dashboard.rs`. lmk if it's not okay with you.

# Testing
* The existing Pytest harness continues to pass.
* I verified my code using the example publisher and the `integration-tests` config (that's also how I generated the screenshot). The timestamps and price values update often, as expected. The rightmost "last local update time" column displays "no data" unil updates start flowing in from example publisher, as expected. In the screenshot you can notice the column contains a newer timestamp, demonstrating that the local data is more up to date     ...as expected.

# Review Highlights
* `metrics.rs` - I am not  100% happy with the readability of the code that generates the HTML and does lookups in the local/global stores. As I move this code to its own module, It will probably make more sense.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,8 +39,6 @@ humantime-serde = "1.1.1"
 slog-envlogger = "2.2.0"
 serde-this-or-that = "0.4.0"
 typed-html = { git = "https://github.com/bodil/typed-html", rev = "4c13ecca" }
-prometheus = "0.13.3"
-lazy_static = "1.4.0"
 
 [dev-dependencies]
 tokio-util = { version = "0.7.0", features = ["full"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,9 @@ clap = { version = "4.0.32", features = ["derive"] }
 humantime-serde = "1.1.1"
 slog-envlogger = "2.2.0"
 serde-this-or-that = "0.4.0"
+typed-html = { git = "https://github.com/bodil/typed-html", rev = "4c13ecca" }
+prometheus = "0.13.3"
+lazy_static = "1.4.0"
 
 [dev-dependencies]
 tokio-util = { version = "0.7.0", features = ["full"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,12 @@ clap = { version = "4.0.32", features = ["derive"] }
 humantime-serde = "1.1.1"
 slog-envlogger = "2.2.0"
 serde-this-or-that = "0.4.0"
+# The public typed-html 0.2.2 release is causing a recursion limit
+# error that cannot be fixed from outside the crate.
+#
+# Rationale, 2023-03-21: https://stackoverflow.com/questions/74462753
 typed-html = { git = "https://github.com/bodil/typed-html", rev = "4c13ecca" }
+humantime = "2.1.0"
 
 [dev-dependencies]
 tokio-util = { version = "0.7.0", features = ["full"] }
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -1,3 +1,5 @@
+use std::net::SocketAddr;
+
 /* ###################################################### System Architecture #######################################################
 
 +-----------------------------+      +----------------------------+
@@ -46,6 +48,7 @@ Note that there is an Oracle and Exporter for each network, but only one Local S
 
 ################################################################################################################################## */
 
+pub mod metrics;
 pub mod pythd;
 pub mod solana;
 pub mod store;
@@ -81,6 +84,7 @@ impl Agent {
     }
 
     async fn spawn(&self, logger: Logger) -> Result<()> {
+        // job handles
         let mut jhs = vec![];
 
         // Create the channels
@@ -132,8 +136,8 @@ impl Agent {
         jhs.push(pythd::adapter::spawn_adapter(
             self.config.pythd_adapter.clone(),
             pythd_adapter_rx,
-            global_store_lookup_tx,
-            local_store_tx,
+            global_store_lookup_tx.clone(),
+            local_store_tx.clone(),
             shutdown_tx.subscribe(),
             logger.clone(),
         ));
@@ -143,8 +147,16 @@ impl Agent {
             self.config.pythd_api_server.clone(),
             pythd_adapter_tx,
             shutdown_rx,
-            logger,
+            logger.clone(),
         ));
+
+        // Spawn the metrics server
+        jhs.push(tokio::spawn(metrics::MetricsServer::spawn(
+            "127.0.0.1:8888".parse::<SocketAddr>()?,
+            local_store_tx,
+            global_store_lookup_tx,
+            logger,
+        )));
 
         // Wait for all tasks to complete
         join_all(jhs).await;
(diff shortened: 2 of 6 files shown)


src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![recursion_limit = "256"]
```
ali-bahjati, line 1: What is the reason for this?

Cargo.toml
```diff
@@ -38,6 +38,7 @@ clap = { version = "4.0.32", features = ["derive"] }
 humantime-serde = "1.1.1"
 slog-envlogger = "2.2.0"
 serde-this-or-that = "0.4.0"
+typed-html = { git = "https://github.com/bodil/typed-html", rev = "4c13ecca" }
```
ali-bahjati, line 41: Do you mind commenting why you are using this as a git dependency?

src/agent.rs
```diff
@@ -143,9 +147,17 @@ impl Agent {
             self.config.pythd_api_server.clone(),
             pythd_adapter_tx,
             shutdown_rx,
-            logger,
+            logger.clone(),
         ));
 
+        // Spawn the metrics server
+        jhs.push(tokio::spawn(metrics::MetricsServer::spawn(
+            "127.0.0.1:8888".parse::<SocketAddr>()?,
```
ali-bahjati, line 155: I think it's better to have a metric config and this be the default value for it.

src/agent.rs
```diff
@@ -1,3 +1,5 @@
+use std::net::SocketAddr;
```
ali-bahjati, line 1: nitpik: I think it's better to move this down.

src/agent/metrics.rs
```diff
@@ -0,0 +1,376 @@
...
+        uptime_seconds %= 3600;
+
+        let uptime_minutes = uptime_seconds / 60; // 1min
+        uptime_seconds %= 60;
+
+        let uptime_string = format!(
+            "{}d{}h{}m{}s",
+            uptime_days, uptime_hours, uptime_minutes, uptime_seconds
+        );
+
```
ali-bahjati, line 158: You can use use `humantime::format_duration` to just get the same thing:

```rust
use humantime::format_duration;
let duration_string = format_duration(duration);
```

src/agent/metrics.rs
```diff
@@ -0,0 +1,376 @@
...
+///
+/// The view is indexed by human-readable symbol name or a stringified
+/// public key if symbol name can't be found.
+pub fn build_dashboard_data(
+    mut local_data: HashMap<PriceIdentifier, PriceInfo>,
+    mut global_data: AllAccountsData,
+    mut global_metadata: AllAccountsMetadata,
+    logger: &Logger,
+) -> BTreeMap<String, DashboardSymbolView> {
+    let mut ret = BTreeMap::new();
```
ali-bahjati, line 275: Is there any reason that it's a map?

src/agent/metrics.rs
```diff
@@ -0,0 +1,376 @@
...
+        Reply,
+    },
+};
+
+/// Internal metrics server state, holds state needed for serving
+/// dashboard and metrics.
+pub struct MetricsServer {
+    /// Used to pull the state of all symbols in local store
+    local_store_tx:         mpsc::Sender<Message>,
+    global_store_lookup_tx: mpsc::Sender<Lookup>,
```
jayantk, line 59: i presume in the future you're going to create another component that aggregates metrics over time (like gets sent info on each update or whatever to track in accumulators), and the data from that component will be accessible here as well?

I think the current information here is a good start to a dashboard, but I'm pretty sure that people will want to see metrics over some time interval. Most problems (like dropped RPC requests, etc) will happen sporadically and you'll want historical data to see whether or not they've been happening.

jayantk (commented):
looks good to me. I agree with ali's comments too, so i'll let him approve when you've addressed them.

src/agent/metrics.rs
```diff
@@ -0,0 +1,376 @@
...
+                    if let Some(datetime) =
+                        NaiveDateTime::from_timestamp_opt(local_data.timestamp, 0)
+                    {
+                        datetime.format("%Y-%m-%d %H:%M:%S").to_string()
+                    } else {
+                        format!("Invalid timestamp {}", local_data.timestamp)
+                    }
+                } else {
+                    "no data".to_string()
+                };
```
ali-bahjati, line 194: nitpik: maybe in the future I move these to another method to separate the html rendering vs what data to display.

ali-bahjati (approved):
This looks good. I think it's good to wait a bit for Tom to review this too.
I left some comments and appreciate if you address them before merging.

src/agent/metrics.rs
drozdziak1: BTree* family containers automatically sort their keys as opposed to the regular Hash* implementations. The goal was to get a sorted view every time

src/agent/metrics.rs
drozdziak1: The main piece of Rust boilerplate for adding metrics is that the Rust Prometheus crate does not let you discover existing metrics easily from global metrics state. This state is kept in a `Registry` object, which is the top-level container used to gather and export formatted metrics in bulk. The preferred method for creating/referencing metrics is linking counter/gauge objects to the registry, and using the objects themselves as references to individual metrics, presumably right beside other state variables. 

This pattern was reasonably easy for the attester, where we focused on the overall performance of the system and got away with storing a handful of static metrics as globals.

In the agent, we'll likely need to add the counter objects per-symbol dynamically. I plan to extend global/local store state to hold the new objects and update them with incoming Update* messages. I'ii take good a look at Richard's comments and screenshots to get an idea of what could work well.

src/agent/metrics.rs
drozdziak1: The primary reason I used a map was deduplication. Come to think of it, I should probably detect overwrites on existing entries

src/agent/metrics.rs
drozdziak1: I already broke ground on the metrics front, starting with rearranging the modules. It definitely looks better that way and is coming soon.

src/agent/metrics.rs
drozdziak1: Done 🫡

src/agent.rs
drozdziak1: Done 🫡

src/agent.rs
drozdziak1: Done 🫡

src/lib.rs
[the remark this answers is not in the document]
drozdziak1: Done 🫡

PR #51, opened 2023-03-22
Add Remote Keypair Loader, a new keypair hotloading mechanism

drozdziak1:
# How it works
We open two POST endpoints, `/primary/load_keypair` and `secondary/load_keypair` on a configurable address and port. Each takes a regular Solana keypair in json (`[1, 241, 24, [...], 3]`).  We check that the account holds the configured minimum SOL balance and make it available in the system if the check passes. The endpoints only have an effect if a publisher keypair is not a valid file path on startup. This can be achieved by specifying a name like `""`. A warning is printed in logs when that happens, informing about the service waiting for the key to that network. If the keypair _is_ valid, it is used at all times like before, ignoring the remote key loader completely.

# Complexity
This new system lets you change the key of either network as many times as you wish. This has complicated the design noticeably, in comparison to something like a busy loop waiting for a single upload request before instantiating other components of the system. Here's my counter-points:
* If an admin uploads the wrong key, they'll need to restart the service, possibly crippling the uptime on the other network (e.g. they botch primary network upload, while secondary is working fine)
* If we ever split the keypairs into separate payer and identity (good practice IMO), swapping the payer mid flight will be much easier to implement, because the  key change notification code will be already there. 

# Review highlights
* `exporter.rs` - this is where the uploaded keypair is used for signing and it's important that all uses of `publish_keypair` subscribe to the key changes using the Rust channel that is passed there. It appeared to be used in only one place - the `publish_batch` method.

# Testing
* Verified manually, injecting the keypair during the integration test run, with publisher keypair renamed
* Added a new integration test, responsible for running a publisher with no keypair file and supplying the keypair through the loader before reusing the existing test scenario. Making this a separate test ensures that both modes work correctly.

### DIFF
--- a/src/agent/remote_keypair_loader.rs
+++ b/src/agent/remote_keypair_loader.rs
@@ -1,3 +1,6 @@
+//! Remote keypair loading endpoint. Lets you hotload a keypair in
+//! runtime for publishing to the given network.
+//!
 use {
     anyhow::{
         Context,
--- a/src/agent/remote_keypair_loader.rs
+++ b/src/agent/remote_keypair_loader.rs
@@ -12,7 +12,10 @@ use {
         signer::Signer,
     },
     std::{
-        net::SocketAddr,
+        net::{
+            IpAddr,
+            SocketAddr,
+        },
         sync::Arc,
         time::Duration,
     },
@@ -40,18 +43,16 @@ pub fn default_min_keypair_balance_sol() -> u64 {
 }
 
 pub fn default_bind_address() -> SocketAddr {
-    "0.0.0.0:9001"
+    "localhost:9001"
         .parse()
         .expect("INTERNAL: Could not build default remote keypair loader bind address")
 }
 
 #[derive(Clone, Debug, Deserialize)]
+#[serde(default)]
 pub struct Config {
-    #[serde(default = "default_min_keypair_balance_sol")]
     primary_min_keypair_balance_sol:   u64,
-    #[serde(default = "default_min_keypair_balance_sol")]
     secondary_min_keypair_balance_sol: u64,
-    #[serde(default = "default_bind_address")]
     bind_address:                      SocketAddr,
 }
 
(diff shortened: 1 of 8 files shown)


src/agent/remote_keypair_loader.rs
```diff
@@ -0,0 +1,318 @@
...
+    mut secondary_rx: mpsc::Receiver<KeypairRequest>,
+    shared_state: Arc<Mutex<RemoteKeypairLoader>>,
+    logger: Logger,
+) {
+    loop {
+        let locked_state = shared_state.lock().await;
+
+        // Only handle requests for defined keypairs. The possibility
+        // of missing keypair is the reason we are not
+        // tokio::select!()-ing on the two channel receivers.
```
ali-bahjati, line 283: I guess maybe another way could be dropping them as they come? 🤔

src/agent/solana/exporter.rs
```diff
@@ -321,6 +334,21 @@ impl Exporter {
...
+        } else {
+            debug!(
+                self.logger,
+                "Exporter: Publish keypair is None, requesting remote loaded key"
+            );
+            let kp = RemoteKeypairLoader::request_keypair(&self.keypair_request_tx).await?;
+            debug!(self.logger, "Exporter: Keypair received");
+            kp
+        };
+
```
ali-bahjati, line 351: I think it is better to drop the request if it cannot get the key because when in the current code when it gets the key the api will bloat the rpc.

ali-bahjati (commented):
The code is very well written and it's great that you added it to the integration tests 🔥 . I think as suggested before, changing the default to localhost and giving enough warning on the config.toml doc is good. (in general it's good to add the configs to the config.toml as the place that explains things with their default value) 

Apart from that, your design choice to allow changing keys is interesting but it's a little complex. I left a comment and would appreciate if you handle it. I don't have any strong objection about the design but I recommend commenting it briefly on the code (a summary of what you already put on the PR description)

PR #53, opened 2023-03-27
oracle.rs: Account -> Entry for clarity, lookup mapping with fewer requests

drozdziak1:
Some of our publishers are concerned about the amount of file descriptors used by the agent process, which could cause problems as the symbol set on chain grows.

# How this works
We used to look up mapping/product/price accounts using individual Solana RPC requests, opening concurrently a TCP connection per product and another one for each of its prices,

This change replaces these individual product lookups with a single `get_multiple_accounts` call, optionally dividing the bwork into chunks with a new option, `max_lookup_batch_size`. Price lookups are now done in a similar way, by first looking up every first price of each product in a single request, then all second prices (if found) and so on. 

# Testing
* Tests continue to pass locally - To me this suggests no obvious errors, feel free to suggest any edge cases we might want to look at.

# Review highlights
* `oracle.rs` - The price lookup logic (around line 500) introduces most complexity. It's important that the new logic is correct, easy to understand and that the complexity is justified. I don't think it's very hard, but as author I'm biased.

### DIFF
--- a/src/agent/solana/oracle.rs
+++ b/src/agent/solana/oracle.rs
@@ -103,6 +103,13 @@ pub struct Config {
     pub updates_channel_capacity: usize,
     /// Capacity of the channel over which the Poller sends data to the Oracle
     pub data_channel_capacity:    usize,
+
+    /// Ask the RPC for up to this many product/price accounts in a
+    /// single request. Tune this setting if you're experiencing
+    /// timeouts on data fetching. In order to keep concurrent open
+    /// socket count at bay, the batches are looked up sequentially,
+    /// trading off overall time it takes to fetch all symbols.
+    pub max_lookup_batch_size: usize,
 }
 
 impl Default for Config {
@@ -113,6 +120,7 @@ impl Default for Config {
             subscriber_enabled:       true,
             updates_channel_capacity: 10000,
             data_channel_capacity:    10000,
+            max_lookup_batch_size:    200,
         }
     }
 }
@@ -152,6 +160,7 @@ pub fn spawn_oracle(
         rpc_timeout,
         config.commitment,
         config.poll_interval_duration,
+        config.max_lookup_batch_size,
         logger.clone(),
     );
     jhs.push(tokio::spawn(async move { poller.run().await }));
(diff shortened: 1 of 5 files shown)


src/agent/solana/oracle.rs
```diff
@@ -419,88 +417,102 @@ impl Poller {
...
+        let mut price_entries = HashMap::new();
 
-        Ok((product_account, price_accounts))
-    }
+        // Starting with top-level prices, look up price accounts in
+        // batches, filling price entries and adding found prices to
+        // the product entries
+        let mut todo = product_entries
+            .iter()
+            .map(|(_key, p)| p.account_data.px_acc)
```
ali-bahjati, line 471: nitpik: based on clippy you can use `product_entires.values().map()`

src/agent/solana/oracle.rs
```diff
@@ -419,88 +417,102 @@ impl Poller {
...
-            .flat_map(|(_, price_accounts)| price_accounts.into_iter())
-            .collect();
-
-        Ok((product_accounts, price_accounts))
-    }
+        // Look up all products with a single request
+        let product_accounts = self
+            .rpc_client
+            .get_multiple_accounts(product_keys.as_slice())
+            .await?;
```
ali-bahjati, line 441: I am a bit worried that RPCs cannot handle this properly. Wdyt about batching it?

https://github.com/pyth-network/pyth-agent/actions/workflows/image-push.yaml 
Here we can dispatch a workflow from branch and with this I can test the new version and make sure this works.

ali-bahjati (commented):
Very nice. I like renaming Price/Product Account to Entry and it makes things clear. My only worry is that the RPCs cannot handle 100+ account or they time out. I think we need to batch them. We can create a release and test it in devnet/testnet without pushing it to be sure. Can you also bump the version?

src/agent/solana/oracle.rs
drozdziak1: I think this is a good idea, I will add batching these requests. Users can always set this to an absurdly high value if their RPC can handle it.

src/agent/solana/oracle.rs
drozdziak1: Nice, thanks!

src/agent/solana/oracle.rs
drozdziak1: Done 🫡

ali-bahjati (approved):
🚢 it ser

PR #54, opened 2023-03-28
Prometheus metrics

drozdziak1:
This change introduces an initial set of Prometheus metrics:
* Selection of PriceInfo fields from local store
* Selection of PriceAccount fields from global store
* Selection of ProductAccount fields from global store

The metrics are constructed dynamically, depending on the symbols existing in global/local store. Relevant symbol names and pubkeys are marked using labels.

# Testing
* Tests continue to pass locally
* The full set of metrics is shown when curling the metrics endpoint (localhost:8888/metrics)

# Review Highlights
* `metrics.rs` - this diff includes moving the simple dashboard do its own module in `dashboard.rs`. Remaining volume of the diff is mostly repetitive registering and updating of the metrics objects. We guard  that all metrics are used with destructuring with errors on unused variable warnings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1008,6 +1008,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "dtoa"
+version = "1.0.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "65d09067bfacaa79114679b279d7f5885b53295b1e2cfb4e79c8e4bd3d633169"
+
+[[package]]
 name = "dyn-clone"
 version = "1.0.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


ali-bahjati (commented):
This is implement very well and I like you refactoring along the way. I like to discuss the design more with you before approving it. 

Since the whole code base is designed around actor model for more resiliency I was thinking maybe it's better if we make the metrics server another actor and other actors communicate with it instead of storing it. Of course since many actors will use it it might not be ideal but I like to hear your opinions. 

Also, I think with this new metric model, the dashboard might be able to use the data from the metrics server (either the prom metrics themself or a local storage) instead of sending messages to the local global store. This will make the code about the dashboard easier to understand.

ali-bahjati (approved):

PR #58, opened 2023-04-06
Add sample configs

ali-behjati:


config/config.sample.pythnet.toml
```diff
@@ -0,0 +1,39 @@
...
+exporter.publish_interval_duration = "400ms"
+
+# Configuration for the optional secondary network this agent will publish data to.
+# In most cases this should be a Solana endpoint.
+[secondary_network]
+
+# Please use other endpoints as these are rate limited
+rpc_url = "https://api.mainnet-beta.solana.com"
+wss_url = "wss://api.mainnet-beta.solana.com"
+
```
jayantk, line 27: i take it the default exporter interval is 1s so that's why it doesn't need to be here?

jayantk (approved):
very nice thank you

config/config.sample.pythnet.toml
ali-bahjati: yes that's right. I'll add the default value so it be more clear for people.

PR #65, opened 2023-05-04
Drozdziak1/permissioned price accounts

drozdziak1:
This change prevents the agent from publishing prices that are not permissioned to its publishing keypair, potentially limiting the burn rate of the publishers' production deployments. The invalid attempts are logged to WARN.

# How this works
*`solana.rs` - we open a channel between oracle and exporter, periodically updating the latest known set of permissioned price accounts.
* `oracle.rs` - In addition to on-chain state lookup, we filter out permissioned prices and send them over the channel to the exporter. This is done by checking the `PriceComp.publisher` field against the publish keypair from key store. 
* `exporter.rs` - Just before building price update Solana transactions, we check for permissioned price set updates on the channel. If no updates are available, pre-existing value from exporter state is used. We filter out unpermissioned symbols using this set and log them to WARN. We continue with publishing the remaining valid symbols.

# Testing
* I verified manually that the debug-level log output corresponds with the desired unpermissioned price warning in a temporary integration test.
* A new integration test is added which verifies that unpermissioned symbols logging works and does not produce transactions

# Review highlights
* Hangups and channel overflows - by creating a new async channel, we risk bugs related to empty/overflowing channels. I did my best to fall back on empty channel and never let it overflow, but you can never be too sure.

# Remaining items
* [x] Proper Integration Testing - I'm still not 100% sure how to best add this to the integration tests. Because of our choice to use an infallible tx, our only choice seems to be parsing agent or solana tx logs. I don't like this, but if we don't come up with something else, that's probably what I'll do

### DIFF
--- a/integration-tests/tests/test_integration.py
+++ b/integration-tests/tests/test_integration.py
@@ -78,7 +78,20 @@ AAPL_USD = {
     },
     "metadata": {"jump_id": "186", "jump_symbol": "AAPL", "price_exp": -5, "min_publishers": 1},
 }
-ALL_PRODUCTS=[BTC_USD, AAPL_USD]
+
+ETH_USD = {
+    "account": "",
+    "attr_dict": {
+        "symbol": "Crypto.ETH/USD",
+        "asset_type": "Crypto",
+        "base": "ETH",
+        "quote_currency": "USD",
+        "generic_symbol": "ETHUSD",
+        "description": "ETH/USD",
+    },
+    "metadata": {"jump_id": "78876710", "jump_symbol": "ETHUSD", "price_exp": -8, "min_publishers": 1},
+}
+ALL_PRODUCTS=[BTC_USD, AAPL_USD, ETH_USD]
 
 asyncio.set_event_loop(asyncio.new_event_loop())
 
@@ -237,7 +250,7 @@ class PythTest:
     def refdata_products(self, refdata_path):
         path = os.path.join(refdata_path, 'products.json')
         with open(path, 'w') as f:
-            f.write(json.dumps([BTC_USD, AAPL_USD]))
+            f.write(json.dumps(ALL_PRODUCTS))
             f.flush()
             yield f.name
 
@@ -257,6 +270,7 @@ class PythTest:
             f.write(json.dumps({
                     "AAPL": {"price": ["some_publisher"]},
                     "BTCUSD": {"price": ["some_publisher"]},
+                    "ETHUSD": {"price": []},
                     }))
             f.flush()
             yield f.name
@@ -430,24 +444,27 @@ class PythTest:
             yield
 
     @pytest_asyncio.fixture
-    async def client(self, agent):
+    async def client(self, agent, tmp_path):
         client = PythAgentClient(address="ws://localhost:8910")
         await client.connect()
-        yield client
+        yield (client, tmp_path)
         await client.close()
 
     @pytest_asyncio.fixture
     async def client_hotload(self, agent_hotload):
         client = PythAgentClient(address="ws://localhost:8910")
         await client.connect()
-        yield client
+        yield (client, tmp_path)
         await client.close()
 
 
 class TestUpdatePrice(PythTest):
 
     @pytest.mark.asyncio
-    async def test_update_price_simple(self, client: PythAgentClient):
+    async def test_update_price_simple(self, client: (PythAgentClient, str)):
+
+        (client, tmp_path) = client
+
         # Fetch all products
         products = {product["attr_dict"]["symbol"]: product for product in await client.get_all_products()}
 
@@ -504,7 +521,8 @@ class TestUpdatePrice(PythTest):
             assert conf == 2
 
     @pytest.mark.asyncio
-    async def test_update_price_simple_with_keypair_hotload(self, client_hotload: PythAgentClient):
+    async def test_update_price_simple_with_keypair_hotload(self, client_hotload: (PythAgentClient, str)):
+
         # Hotload the keypair into running agent
         hl_request = requests.post("http://localhost:9001/primary/load_keypair", json=PUBLISHER_KEYPAIR)
 
@@ -517,11 +535,10 @@ class TestUpdatePrice(PythTest):
         await self.test_update_price_simple(client_hotload)
 
     @pytest.mark.asyncio
-    @pytest.mark.skip(reason="Test not meant for automatic CI")
-    async def test_publish_forever(self, client: PythAgentClient):
-        '''
-        Convenience test routine for manual experiments on a running test setup.
-        '''
+    async def test_update_price_discards_unpermissioned(self, client: (PythAgentClient, str)):
+
+        (client, tmp_path) = client
+
         # Fetch all products
         products = {product["attr_dict"]["symbol"]: product for product in await client.get_all_products()}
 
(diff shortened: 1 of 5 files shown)


integration-tests/tests/test_integration.py
```diff
@@ -430,24 +444,27 @@ def agent_hotload(self, sync_accounts, agent_keystore, agent_keystore_path, tmp_
             yield
 
     @pytest_asyncio.fixture
-    async def client(self, agent):
+    async def client(self, agent, tmp_path):
```
jayantk, line 447: why do you need to pass tmp_path through here? seems like this function doesn't use it. I think you could add tmp_path as an argument directly to the tests that need it below

integration-tests/tests/test_integration.py
```diff
@@ -532,11 +623,16 @@ async def test_publish_forever(self, client: PythAgentClient):
...
+        price_account_unperm = product_unperm["price_accounts"][0]["account"]
+
         while True:
             # Send an "update_price" request
             await client.update_price(price_account, 47, 2, "trading")
             time.sleep(1)
 
             # Send another "update_price" request to trigger aggregation
-            await client.update_price(price_account, 81, 1, "trading")
+            await client.update_price(price_account_unperm, 81, 1, "trading")
```
jayantk, line 652: do you think this is the right behavior for the infinite looping test? I haven't thought through what's most useful for this test case, and i could see the warnings being annoying. but maybe there's a good reason to have the unpermissioned symbols here too.

src/agent/solana/exporter.rs
```diff
@@ -285,12 +302,63 @@ impl Exporter {
             })
             .collect::<Vec<_>>();
 
-        if fresh_updates.is_empty() {
+        // Update permissioned accounts of this publisher from the
+        // oracle. The loop ensures that we clear the channel and use
+        // only the final, latest message; try_recv() is non-blocking.
+        loop {
```
jayantk, line 308: I think it would be better to put this in `run` above? I think you can probably set up the publish_interval timer such that this processing happens between the publish_updates calls and doesn't interfere with them.

(also would be better from a code clarity perspective to put this in a separate function like `flush_perm_price_channel()` and call it from run())

integration-tests/tests/test_integration.py
```diff
@@ -430,24 +444,27 @@ def agent_hotload(self, sync_accounts, agent_keystore, agent_keystore_path, tmp_
...
-        yield client
+        yield (client, tmp_path)
         await client.close()
 
     @pytest_asyncio.fixture
     async def client_hotload(self, agent_hotload):
         client = PythAgentClient(address="ws://localhost:8910")
         await client.connect()
-        yield client
+        yield (client, tmp_path)
```
ali-bahjati, line 457: couldn't we add another fixture for the path specifically? It doesn't look very clean like this 🤔

src/agent/solana/exporter.rs
```diff
@@ -285,12 +302,63 @@ impl Exporter {
...
+                if self.perm_price_accounts.contains(&key_from_id) {
+                    true
+                } else {
+                    warn!(self.logger, "Exporter: Attempted to publish a price without permission, skipping";
+			  "price_account" => key_from_id.to_string(),
+			  "perm_price_accounts" => format!("{:?}", self.perm_price_accounts),);
+                    false
+                }
+            })
+            .collect::<Vec<_>>();
```
jayantk, line 354: isn't this going to generate a ton of warnings in normal usage? If every publisher has different permissions on mainnet and pythnet, then I think `fresh_updates` here will contain the union of both. Which means that every time this code runs for solana mainnet, they will get a lot of warnings for the unpermissioned symbols.

src/agent/solana/exporter.rs
```diff
@@ -213,6 +221,12 @@ pub struct Exporter {
     // Channel on which to send inflight transactions to the transaction monitor
     inflight_transactions_tx: Sender<Signature>,
 
+    /// Permissioned symbols as read by the oracle module
+    perm_price_rx: mpsc::Receiver<HashSet<Pubkey>>,
+
+    /// Symbols we are permissioned to publish
+    perm_price_accounts: HashSet<Pubkey>,
```
jayantk, line 228: maybe these permissions should go in `NetworkState`? I think that already contains network-specific information, which this is also.

integration-tests/tests/test_integration.py
```diff
@@ -516,12 +534,85 @@ async def test_update_price_simple_with_keypair_hotload(self, client_hotload: Py
...
+        final_price_account_unperm = final_product_state_unperm["price_accounts"][0]
+        assert final_price_account_unperm["price"] == 0
+        assert final_price_account_unperm["conf"] == 0
+        assert final_price_account_unperm["status"] == "unknown"
+
+        # Confirm agent logs contain the relevant WARN log
+        with open(f"{tmp_path}/agent_logs/stdout") as f:
+            contents = f.read()
+            lines_found = 0
+            for line in contents.splitlines():
```
ali-bahjati, line 608: nitpik: you can do `for line in f` to be more efficient

src/agent/solana/exporter.rs
jayantk: oh also you may want to look into `watch::Receiver` thing for `NetworkState` above, which seems to do exactly the thing you're doing here.

src/agent/solana.rs
```diff
@@ -75,14 +75,19 @@ pub mod network {
         keypair_request_tx: mpsc::Sender<KeypairRequest>,
         logger: Logger,
     ) -> Result<Vec<JoinHandle<()>>> {
+        // Permissioned price account updates between oracle and exporter
+        let (perm_price_tx, perm_price_rx) = mpsc::channel(config.oracle.updates_channel_capacity);
```
ali-bahjati, line 79: just to confirm, is this config ok here?

src/agent/solana/exporter.rs
```diff
@@ -285,12 +302,63 @@ impl Exporter {
...
+
+        // Filter out price accounts we're not permissioned to update
+        let permissioned_updates = fresh_updates
+            .into_iter()
+            .filter(|(id, _data)| {
+                let key_from_id = Pubkey::new(id.clone().to_bytes().as_slice());
+                if self.perm_price_accounts.contains(&key_from_id) {
+                    true
+                } else {
+                    warn!(self.logger, "Exporter: Attempted to publish a price without permission, skipping";
```
ali-bahjati, line 348: I don't think that a WARN log be good. Mostly because some publishers publish for both mainnet and pythnet so this is expected to happen. Can we make it an INFO log level?

ali-bahjati (approved):
Very nice! My only big comment is only changing the log level. If you agree with it, please address that before merging ser.

src/agent/solana/oracle.rs
```diff
@@ -384,24 +414,73 @@ impl Poller {
...
     async fn poll(&self) -> Result<Data> {
         let mapping_accounts = self
-            .fetch_mapping_accounts(self.mapping_account_key)
+            .fetch_mapping_accounts(self.key_store.mapping_key)
             .await?;
         let (product_accounts, price_accounts) = self
             .fetch_product_and_price_accounts(mapping_accounts.values())
             .await?;
 
+        let publish_keypair = if let Some(kp) = self.key_store.publish_keypair.as_ref() {
```
jayantk, line 440: needing the publisher keypair here feels inelegant. I think maybe a simpler way to do this would be to return the full mapping from price accounts -> permissioned publishers, then check which ones are permissioned in the exporter (where we already have this keypair loading logic)

integration-tests/tests/test_integration.py
drozdziak1: oh, you're right, I should just mention the fixture directly in test signature, thanks!

integration-tests/tests/test_integration.py
drozdziak1: I was hesitant to do that because while this assertion takes place, the agent is still running and appending to the file which could create an infinite loop.

src/agent/solana/exporter.rs
jayantk: eh... maybe not. i see why you added this channel, because you want to get the information from the Oracle, which is already getting the other price account information. Maybe this is fine.

jayantk (commented):
I think this could be simplified a bit. See inline comments

src/agent/solana.rs
drozdziak1: This value felt intuitive, because the permissioned accounts also represent updates coming from the oracle

src/agent/solana/exporter.rs
drozdziak1: I think you're right, thanks

integration-tests/tests/test_integration.py
ali-bahjati: oh smart, you are right.

src/agent/solana/exporter.rs
drozdziak1: Exiled this block to a helper method. I got into some nasty Rust errors when trying to use the `watch` channel. I explain this in detail in the new `update_our_prices` method.

I wouldn't put this call inside `run`, because we would either need to duplicate the keypair request code again, or do it in `run` also. I think the intention with it is to just say "we're doing this one thing periodically".

src/agent/solana/exporter.rs
drozdziak1: switched to `info`

src/agent/solana/oracle.rs
drozdziak1: Done, we get a hashmap in exporter and find relevant entry for the publisher using just the single keypair request `if`.

integration-tests/tests/test_integration.py
drozdziak1: I don't think it hurts to have it, but neither does just valid publishing, removed.

jayantk (approved):
very nice

PR #68, opened 2023-05-11
Fix accumulator config name

ali-bahjati:


drozdziak1 (approved):
I missed the correct, top-level config option name when documenting, sorry! Thank you!

PR #69, opened 2023-05-17
Bump Pyth SDK, support JSON-RPC batch requests

drozdziak1:
This change contains Ali's fix for the strange huge memory allocation bug we've seen from several publishers and fixes a followup issue of our publishing API not supporting batch requests.
# Summary of changes
##  `api.rs` - Most changes happen here
* Centralized sending WebSocket responses in `Connection::handle`. This felt like the right choice to build the whole batch response  and send it all at once, as opposed to separately replying with `[`, each request's result and then `]` which would be less readable and necessary to keep using `send_text` all over the place. The square brackets consistency is also a concern - we shouldn't wrap a one-off response in brackets if the request didn't arrive in brackets.
* Restricted error messages to top-level parsing errors, printing detailed messages in logs.
* Started logging all requests to debug!()
* Fixed wrong pyth-sdk imports (newer SDK moves stuff around)
* Updated unit test assertions, using serde_json::json!() for larger payloads for readability
* Added a unit test for the batch requests
* Added a way to read out the logs for test debugging purposes

## `Cargo.toml` - Bump SDK - With the previous version, mishandling the new "ignored" price status caused an out-of-bounds read that resulted in the large allocation 

# Review highlights
* Batch requests are handled sequentially - I had trouble pleasing the compiler into letting me `join_all` on a bunch of request handling futures, mainly because of uncertainty of the return type - I do not believe any of the requests should take more time than a Solana block time or the configured publishing/lookup intervals
* The error response obfuscation may be a little much, but it's very simple to revert if that's what we want.

# Testing
* All tests continue to pass
* `api.rs` has a new dedicated unit test for a batch request scenario


src/agent/pythd/api.rs
```diff
@@ -304,65 +304,127 @@ pub mod rpc {
...
+                .to_str()
+                .map_err(|_| anyhow!("Could not parse message as text"))?;
+
+            let json_value: Value = serde_json::from_str(s)?;
+            if let Some(array) = json_value.as_array() {
+                // Interpret request as JSON-RPC 2.0 batch if value is an array
+                let mut requests = Vec::with_capacity(array.len());
+                for maybe_request in array {
+                    // parse_request() is the only jrpc parsing method which takes &str
+                    let maybe_request_string = serde_json::to_string(maybe_request)?;
```
jayantk, line 357: this comment is confusing. I think you're explaining why you have to re-stringify the json object (there's only one method for parsing requests), but the text isn't explicit enough. I would lead with something like "It's unfortunate that we have to stringify the json object here, but ... "

src/agent/pythd/api.rs
```diff
@@ -897,14 +951,14 @@ pub mod rpc {
             let received_json = test_client.recv_json().await;
 
             // Check that the result is what we expect
-            let expected_json = r#"{"jsonrpc":"2.0","error":{"code":-32603,"message":"Missing request parameters","data":null},"id":5}"#;
+            let expected_json = r#"{"jsonrpc":"2.0","error":{"code":-32603,"message":"Error handling request. Check pyth-agent logs for details.","data":null},"id":5}"#;
```
jayantk, line 954: I don't like this new error response behavior. The old messages were nicer for the user, because they could debug their requests more easily. I suggest reverting that change.

jayantk (approved):
lgtm, but just a couple minor things

src/agent/pythd/api.rs
drozdziak1: Changed to something along what you propose.

src/agent/pythd/api.rs
drozdziak1: Reverted back to full error messages

PR #73, opened 2023-06-21
Unchanged price publishing optimization

drozdziak1:
This change prevents resends of unchanged price data, limited by a delay threshold (5s by default). After the delay passes, the unchanged price will be published anyway to avoid exclusion from aggregate.  This new behavior should help decrease gas costs.

# Changes
* New option called `<network>.exporter.unchanged_publishing_threshold` (5s by default) - delay after which unchanged updates are allowed.
* Additional filtering rule in `exporter.rs`, using the new option.

# Testing
* Unit and integration tests continue to pass normally.

### DIFF
--- a/src/agent/solana/exporter.rs
+++ b/src/agent/solana/exporter.rs
@@ -24,10 +24,7 @@ use {
         join_all,
     },
     key_store::KeyStore,
-    pyth_sdk::{
-        Identifier,
-        UnixTimestamp,
-    },
+    pyth_sdk::Identifier,
     pyth_sdk_solana::state::PriceStatus,
     serde::{
         Deserialize,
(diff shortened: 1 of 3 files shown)


src/agent/solana/exporter.rs
```diff
@@ -348,7 +374,7 @@ impl Exporter {
                     // Note: This message is not an error. Some
                     // publishers have different permissions on
                     // primary/secondary networks
-                    info!(
+                    debug!(
```
drozdziak1, line 377: We've seen complaints about these messages being too frequent. I forgot to mention it in my description. Sorry!

jayantk: thanks for commenting on the PR with context. definitely helps me understand why this changed.

jayantk (approved):
nicely done

PR #74, opened 2023-06-22
Drozdziak1/global store staleness log level

drozdziak1:
Small change to avoid publisher confusion. We've seen complaints about large quantities of this message, we move it do debug to silence it by default.


drozdziak1:
Forgot the rebase, change is still very small.

ali-bahjati (approved):

PR #77, opened 2023-07-17
Literal pubkeys in agent config, v2.0.0

drozdziak1:
# Motivation
The legacy pythd config scheme, using individual files for storing program/mapping/accumulator public keys, is inconvenient and its defaults are misleading. This change replaces the key store root path and individual files with inlined base-58 pubkeys and a single path to reach the publishing keypair. Additionally, the inlined string keys make it much more convenient to distribute complete sample configs to publishers.

# Summary of changes
* Bumped package version to `v2.0.0` - this is a breaking change that will prevent users from running the service without updating their config for the new schema.
* solana.rs: Changed `KeyStore` options to use base58 strings instead of file paths
* `src/bin/agent_migrate_config.rs` - Implemented a small migration tool that will find and inline the file-based pubkeys in an existing config, making it work with the breaking change.
* README.md: Provided migration instructions for both `agent-migrate-config` and updating by hand.
* integration-tests/ - Updated the test fixtures to prepare an up-to-date config, added a test confirming that `agent-migrate-config` works
* config/ - Updated default and sample configs

# Review Highlights
* I've manually updated the sample configs to contain pubkeys relevant to each network. It is important that they are correct.
* I wrote a section about keystore migration in `README.md`. It is important that the description is easy to follow.

# Testing
* Tests continue to pass after updating them to build a config with the new schema
* `agent-migrate-config` is tested by converting a legacy config and using it to perform the simple test case

### DIFF
--- a/config/config.sample.pythtest.toml
+++ b/config/config.sample.pythtest.toml
@@ -40,10 +40,10 @@ wss_url = "wss://api.testnet.solana.com"
 key_store.publish_keypair_path = "/path/to/keypair.json"
 
 # Oracle program pubkey
-key_store.program_key = "8tfDNiaEyrV6Q1U4DEXrEigs9DoDtkugzFbybENEbCDz" # conformance
+key_store.program_key = "8tfDNiaEyrV6Q1U4DEXrEigs9DoDtkugzFbybENEbCDz"
 
 # Oracle mapping pubkey
-key_store.mapping_key = "AFmdnt9ng1uVxqCmqwQJDAYC5cKTkw8gJKSM5PnzuF6z" # conformance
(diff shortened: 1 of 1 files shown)


src/agent/solana.rs
```diff
@@ -181,41 +183,59 @@ mod key_store {
...
-                    .context("reading mapping key")?,
-                accumulator_key,
+                program_key: config.program_key,
+                mapping_key: config.mapping_key,
+                accumulator_key: config.accumulator_key,
             })
         }
+    }
+
+    // Helper methods for strinigified SOL addresses
```
jayantk, line 205: typo: stringified

jayantk (approved):
very nicely done

PR #78, opened 2023-07-20
v2.0.0 nitpicks

drozdziak1:
This change contains small fixes to sample pythnet config, the general `config.toml` and fixes confusing `oracle.rs` log messages. Additionally, package version is bumped to `v2.0.1`

### DIFF
--- a/config/config.toml
+++ b/config/config.toml
@@ -23,7 +23,7 @@ wss_url = "wss://api.pythtest.pyth.network"
 # non-existent file path, the system expects a keypair to be loaded
 # via the remote keypair loader. If the path is valid, the remote
 # keypair loading is disabled.
-key_store.publish_keypair = "/path/to/keypair.json"
+key_store.publish_keypair_path = "/path/to/keypair.json"
 
 # Public key of the oracle program
 key_store.program_key = "RelevantOracleProgramAddress"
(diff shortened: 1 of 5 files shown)


config/config.toml
```diff
@@ -23,83 +23,23 @@ wss_url = "wss://api.pythtest.pyth.network"
...
-# Relative path to publisher identity keypair
-# w.r.t. `key_store.root_path`. When the specified file is not found
-# on startup, the relevant primary/secondary network will expect a
-# remote-loaded keypair. See remote_keypair_loader options for
-# details.
-# key_store.publish_keypair_path = "publish_key_pair.json" # I exist, remote loading disabled
-# key_store.publish_keypair_path = "none" # I do not exist, remote loading activated for the network
-
-# Relative path to accumulator message buffer program ID. Setting this
+# Pubkey of accumulator message buffer program ID. Setting this
```
drozdziak1, line 36: The older ordering of these options made additional `exporter.*` and `oracle.*` options fall under the `remote_keypair_loader` sections. This means that if someone were to uncomment all config options and sections, they would not end up with a correct config.

With this change, every config section is clearly marked and additional `{primary|secondary}_network` options exist in the correct spot, still under `[primary_network]`

jayantk (approved):

PR #82, opened 2023-10-12
Update config.sample.pythnet.toml

hkaliraipyth:
added additional note: 

# The pythnet accumulator key settings are only valid for pythnet. Do not apply these settings for any other environment i.e. mainnet, pythtest-conformance


drozdziak1 (commented):
This setting is also valid for `pythtest-crosschain`

hkaliraipyth:
> This setting is also valid for `pythtest-crosschain`

Apart from Jump and the Pyth Back Up publisher - no other publisher will be publishing to pythtest-crosschain, so no need to mention this.

drozdziak1 (approved):

hkaliraipyth:
> This setting is also valid for `pythtest-crosschain`

Apart from the 2 existing publishers, no additional publishers will be publishing to pythtest-crosschain - so no need to note

PR #87, opened 2023-11-13
Use staleness for pricing

ali-bahjati:
This PR adds dynamic (exponential) pricing based on staleness of the feeds that the publishers publish. This feature, combined with the recent fees for the price updates causes the spike on fee to last some time before going away.

The second commit contains a little refactoring to fix warnings. Please review commit-by-commit.

### DIFF
--- a/src/agent/solana/exporter.rs
+++ b/src/agent/solana/exporter.rs
@@ -35,10 +35,8 @@ use {
     solana_client::{
         nonblocking::rpc_client::RpcClient,
         rpc_config::RpcSendTransactionConfig,
-        rpc_response::RpcPrioritizationFee,
     },
     solana_sdk::{
-        bs58,
         commitment_config::CommitmentConfig,
         compute_budget::ComputeBudgetInstruction,
         hash::Hash,
@@ -334,7 +332,7 @@ impl Exporter {
         let permissioned_updates = self.get_permissioned_updates().await?;
         let price_accounts = permissioned_updates
             .iter()
-            .map(|(identifier, _)| Pubkey::new(&identifier.to_bytes()))
+            .map(|(identifier, _)| Pubkey::from(identifier.to_bytes()))
             .collect::<Vec<_>>();
         self.recent_compute_unit_price_micro_lamports = self
             .estimate_compute_unit_price_micro_lamports(&price_accounts)
@@ -470,7 +468,7 @@ impl Exporter {
         Ok(fresh_updates
             .into_iter()
             .filter(|(id, _data)| {
-                let key_from_id = Pubkey::new((*id).clone().to_bytes().as_slice());
+                let key_from_id = Pubkey::from((*id).clone().to_bytes());
                 if self.our_prices.contains(&key_from_id) {
                     true
                 } else {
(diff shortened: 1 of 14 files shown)


src/agent/solana/exporter.rs
```diff
@@ -657,17 +671,65 @@ impl Exporter {
...
+            let maximum_unit_price =
+                MAXIMUM_TOTAL_COMPUTE_UNIT_FEE_MICRO_LAMPORTS / total_compute_limit as u64;
+
+            // Use the estimated previous price if it is higher
+            // than the current price.
+            if let Some(estimated_recent_price) = self.recent_compute_unit_price_micro_lamports {
+                // Get the estimated compute unit price and wrap it so it stays below the maximum total
+                // compute unit fee. We additionally divide such price by 2 to make sure a spike
+                // doesn't get propagated forever.
+                let estimated_price = (estimated_recent_price >> 1).min(maximum_unit_price);
```
m30m, line 684: hardcore division 😎

m30m (approved):

Reisen (approved):

src/agent/solana/exporter.rs
```diff
@@ -657,17 +671,65 @@ impl Exporter {
...
+
+                compute_unit_price_micro_lamports = compute_unit_price_micro_lamports
+                    .map(|price| price.max(estimated_price))
+                    .or(Some(estimated_price));
+            }
+
+            // Use exponentially higher price if this publisher hasn't published in a while for the accounts
+            // in this batch. This will use the maximum total compute unit fee if the publisher
+            // hasn't updated for >= MAXIMUM_SLOT_GAP_FOR_DYNAMIC_COMPUTE_UNIT_PRICE slots.
```
jayantk, line 693: this is pretty much guaranteed to use the maximum price when the market opens each day right?

I think that behavior is fine -- I think the max price isn't so high that we can't afford 1 of these transactions per day per price feed or whatever -- though probably worth a comment justifying it.

src/agent/solana/exporter.rs
```diff
@@ -88,6 +87,7 @@ const UPDATE_PRICE_NO_FAIL_ON_ERROR: i32 = 13;
 // measure while using dynamic compute price to prevent the exporter from paying too much for a
 // single transaction
 const MAXIMUM_TOTAL_COMPUTE_UNIT_FEE_MICRO_LAMPORTS: u64 = 1_000_000_000_000;
+const MAXIMUM_SLOT_GAP_FOR_DYNAMIC_COMPUTE_UNIT_PRICE: u64 = 25;
```
jayantk, line 90: I suggest putting both of these consts into the config. these are the kind of thing that we'll probably want to tune easily in an emergency.

jayantk (approved):
LGTM. left a couple comments but feel free to address in a separate PR

src/agent/solana/exporter.rs
ali-bahjati: Yeah it was a certain challenge. I think we should see how it will work with the other dynamic mechanism (which is decaying) and make sure it won't get very expensive.

PR #91, opened 2023-11-16
fix: use publisher latest slot for staleness based pricing

ali-bahjati:
This PR will use the latest slot of the publisher for staleness pricing. The issues that this solve is a publisher publishing for a not yet live feed and also a publisher publishing during market close hours. We can roll out this change to all the publishers without the market-close change and it should be similarly effective.

This change also includes changing the default variables for staleness gap and max paid price for priority fees.
- staleness gap rationale: the rpc might be a bit behind by a few slots. Considering this delay and the exponential pricing this should change the unit price when the feed is really getting stale and doubling if it is still stale.
- max paid price: the lowered value will result in almost 5m priority fee which is very high.

We will modify them later based on the performance of the publishers.

### DIFF
--- a/src/agent/solana/exporter.rs
+++ b/src/agent/solana/exporter.rs
@@ -154,8 +154,12 @@ impl Default for Config {
             compute_unit_limit:                              40000,
             compute_unit_price_micro_lamports:               None,
             dynamic_compute_unit_pricing_enabled:            false,
-            maximum_total_compute_fee_micro_lamports:        1_000_000_000_000,
-            maximum_slot_gap_for_dynamic_compute_unit_price: 25,
+            // Maximum total compute unit fee paid for a single transaction (0.0001 SOL)
+            maximum_total_compute_fee_micro_lamports:        100_000_000_000,
+            // A publisher update is not included if it is 25 slots behind the current slot.
+            // Due to the delay in the network (until a block gets confirmed) we add 5 slots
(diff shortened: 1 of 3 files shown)


src/agent/solana/exporter.rs
```diff
@@ -722,12 +722,20 @@ impl Exporter {
...
-                .map(|account| account.last_slot)
+                .flat_map(|account| {
+                    account
+                        .comp
+                        .iter()
+                        .filter(|c| c.publisher == publish_keypair.pubkey())
+                        .collect::<Vec<_>>()
+                        .get(0)
+                        .map(|c| c.latest.pub_slot)
+                })
```
Reisen, line 738: Nitpick you can ignore as I know you want to move this faster, but think this is equivelent to just:

```
account
    .comp
    .iter()
    .find(|c| c.publisher == publish_keypair.pubkey())
    .map(|c| c.latest.pub_slot)
```

Are you sure this won't fail with a panic!() in your current code on the `.get(0)`? I trust the assumption is they wouldn't be looking at the feed if not, but just want to confirm?

Reisen (approved):

Reisen:
Can you update the PR text with a little bit more clarity around the numbers, what you expect the burn-rate to be practically in the scenario this covers?

ali-bahjati:
I added some comments on the code.

PR #93, opened 2023-11-21
src/agent/market_hours.rs: Add market hours format implementation

drozdziak1:
# Motivation
This is the first in a series of changes realizing the market hours feature for conditional publishing according to dedicated market hours metadata. This change introduces a market hours format that lets us define a weekly schedule for a market, along with its local timezone. The format works as follows:
```
MarketHours ::= TimeZoneSpec,MHKind,MHKind,MHKind,MHKind,MHKind,MHKind,MHKind
TimeZoneSpec ::= Continent/Country  # e.g. Europe/Paris
MHKind::= O | C | HHMMTime-HHMMTime # e.g. 09:00-17:00, O means Open all day, C means Closed all day
HHMMTime ::= [0-23]:[0-59] # e.g. 15:21
```

## Examples
* 9 - 5 Mon - Fri in Hong Kong `Asia/Hong_Kong,9:00-17:00,9:00-17:00,9:00-17:00,9:00-17:00,9:00-17:00,C,C`
* 24/7 (e.g. many crypto assets) `Europe/Lisbon,O,O,O,O,O,O,O` - This will probably be the default for symbols with no market hours metadata

# Summary of changes
* `src/agent/metrics.rs` - rustfmt really wanted me to add this
* `src/agent/market_hours.rs` - The core of the change
* `Cargo.toml`, `Cargo.lock` - Bump `chrono` version, add `chrono_tz`

# Review highlights
* `src/agent/market_hours.rs` - I think it would make sense to have more non-trivial timezone handling tests, including DST chanes etc.,ideas welcome


src/agent/market_hours.rs
```diff
@@ -0,0 +1,353 @@
...
+}
+
+impl MHKind {
+    pub fn can_publish_at(&self, when_market_local: NaiveTime) -> bool {
+        dbg!(&when_market_local);
+        dbg!(self);
+        match self {
+            Self::Open => true,
+            Self::Closed => false,
+            Self::TimeRange(start, end) => start <= &when_market_local && &when_market_local <= end,
```
jayantk, line 172: I think you need to support 24:00 in the format so that you can have intervals like:

17:00-24:00,00:00-17:00 (for 5pm on the first day through 5pm on the second day).

I think this current check will stop publishing between 11:59 pm and midnight every day if you're not careful.

jayantk: also can you add a test case for this day wraparound behavior?

jayantk (commented):
nice. I think there's one issue here with day wrapping (see inline) but good aside from that.

jayantk (approved):
nice

PR #95, opened 2023-11-24
exporter: Use market hours information in price filtering

drozdziak1:
# Motivation
This change builds on the previously introduced market hours format and applies it to the agent's publishing logic. The extraction, parsing and distribution of the market hours info piggy-backs the existing publisher permissions flow. This felt appropriate because of a shared idea: "Does on-chain data say this is okay to publish?". You'll quickly see that a big part of the changeset is a series of additions to the publisher permissions processing logic. In the same spirit, the changes to state simply add market hours to the detected permissioned symbols that are checked when filtering prices.

# Summary of changes
* `oracle.rs` - Attempt to parse market hours when looking up products, fall back to 24/7 if failing po parse. Specifically, print a warning if the market hours field is defined but could not be understood. Additionally, the market hours are passed together with permissioned prices in the existing channel going to exporter.
* `exporter.rs` - Use market hours data in price filtering.
* `market_hours.rs` - Add missing traits to please the compiler; make `can_publish_at` infallible (turns out the weekday extraction exists in `chrono` but it required some digging)
* `test_integration.py` - Add all-closed market hours to `AAPL_USD` and expect publishing to it to fail because of the market hours.

# Review Highlights
* The new metadata field is named "market_hours"
* The fallback behavior for market hours parsing problems is to use 24/7 market hours. In case we fatfinger the field on a symbol, it feels more appropriate to risk publishing it at a wrong time than to bring it down for all pubs when it's supposed to be up. I think it's important for us to agree about this.
* The point of reference for market hours is the OS system clock - I decided against using the solana clock since it is known to drift from time to time (there were incidents of sizable lag in the past on Solana's public networks)
* `test_integration.py` - the test is very rudimentary, using an all-closed market hours schedule, expecting the symbol to be unchanged. This is mostly due to agent using system clock to determine the current date/time. Fixturing the system clock feels unnecessary when dedicated edge case testing is available in the unit testsuite. The goal of the integration test is to verify that the code path successfully results in no publishing for the all-closed market hours. In case the symbol's hours data were to be misunderstood/lost, it would presumably fall back to a 24/7 schedule, exposing the problems along the code path. If you have good suggestions for making the test less trivial, it would be nice to discuss that.

### DIFF
--- a/integration-tests/tests/test_integration.py
+++ b/integration-tests/tests/test_integration.py
@@ -78,6 +78,7 @@ AAPL_USD = {
         "nasdaq_symbol": "AAPL",
         "symbol": "Equity.US.AAPL/USD",
         "base": "AAPL",
+        "market_hours": "America/New_York,C,C,C,C,C,C,C" # Should never be published due to all-closed market hours
     },
     "metadata": {"jump_id": "186", "jump_symbol": "AAPL", "price_exp": -5, "min_publishers": 1},
 }
@@ -732,3 +733,37 @@ class TestUpdatePrice(PythTest):
             # Continue with the simple test case, which must succeed
             await self.test_update_price_simple(client_no_spawn)
             await client_no_spawn.close()
+
+    @pytest.mark.asyncio
+    async def test_agent_respects_market_hours(self, client: PythAgentClient):
+        '''
+        Similar to test_update_price_simple, but using AAPL_USD and
+        asserting that nothing is published due to the symbol's
+        all-closed market hours.
+        '''
+
+        # Fetch all products
+        products = {product["attr_dict"]["symbol"]: product for product in await client.get_all_products()}
+
+        # Find the product account ID corresponding to the AAPL/USD symbol
+        product = products[AAPL_USD["attr_dict"]["symbol"]]
+        product_account = product["account"]
+
+        # Get the price account with which to send updates
+        price_account = product["price_accounts"][0]["account"]
+
+        # Send an "update_price" request
+        await client.update_price(price_account, 42, 2, "trading")
+        time.sleep(2)
+
+        # Send another "update_price" request to trigger aggregation
+        await client.update_price(price_account, 81, 1, "trading")
+        time.sleep(2)
+
+        # Confirm that the price account has been updated with the values from the first "update_price" request
+        final_product_state = await client.get_product(product_account)
+
+        final_price_account = final_product_state["price_accounts"][0]
+        assert final_price_account["price"] == 0
+        assert final_price_account["conf"] == 0
+        assert final_price_account["status"] == "unknown"
--- a/integration-tests/tests/test_integration.py
+++ b/integration-tests/tests/test_integration.py
@@ -78,7 +78,7 @@ AAPL_USD = {
         "nasdaq_symbol": "AAPL",
         "symbol": "Equity.US.AAPL/USD",
         "base": "AAPL",
-        "market_hours": "America/New_York,C,C,C,C,C,C,C" # Should never be published due to all-closed market hours
+        "weekly_schedule": "America/New_York,C,C,C,C,C,C,C" # Should never be published due to all-closed market hours
     },
     "metadata": {"jump_id": "186", "jump_symbol": "AAPL", "price_exp": -5, "min_publishers": 1},
 }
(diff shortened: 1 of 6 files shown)


integration-tests/tests/test_integration.py
```diff
@@ -732,3 +733,37 @@ async def test_agent_migrate_config(self,
...
             await self.test_update_price_simple(client_no_spawn)
             await client_no_spawn.close()
+
+    @pytest.mark.asyncio
+    async def test_agent_respects_market_hours(self, client: PythAgentClient):
+        '''
+        Similar to test_update_price_simple, but using AAPL_USD and
+        asserting that nothing is published due to the symbol's
+        all-closed market hours.
+        '''
```
jayantk, line 743: Ideally you would be able to write a test that had nontrivial hours (so closed sometimes and open others) and then test both cases where it should / shouldn't publish. I'm not sure how easy that is to do though given that you would need to mock the time for the agent.

src/agent/market_hours.rs
```diff
@@ -53,13 +54,11 @@ impl MarketHours {
         }
     }
 
-    pub fn can_publish_at<Tz: TimeZone>(&self, when: &DateTime<Tz>) -> Result<bool> {
```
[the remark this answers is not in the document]
jayantk: sorry one more request not related to this PR. Can you add some tests for `can_publish_at` that specifically focus on the timezone. like the time X in tz A is ok, but X in tz B is not. 

and maybe also worth testing days with daylight savings time shifts explicitly

jayantk (approved):
seems like a solid solution to me. I left a couple requests for testing but feel free to do in follow up PRs.

src/agent/solana/oracle.rs
```diff
@@ -525,10 +538,29 @@ impl Poller {
                 let product = load_product_account(prod_acc.data.as_slice())
                     .context(format!("Could not parse product account {}", product_key))?;
 
+                let market_hours: MarketHours = if let Some((_mh_key, mh_val)) =
+                    product.iter().find(|(k, _v)| *k == "market_hours")
```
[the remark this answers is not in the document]
drozdziak1: I think I like `weekly_schedule` the best, even my code comment for the `MarketHours` struct kind of points in that direction. I think renaming `MarketHours` to `WeeklySchedule` is also appropriate here. As we tackle holidays in the future, we may choose to create an umbrella `MarketHours` struct with `weekly_schedule: WeeklySchedule` and e.g. `next_holidays: Vec<Holiday>`.

src/agent/solana/exporter.rs
```diff
@@ -468,13 +474,24 @@ impl Exporter {
...
             .into_iter()
             .filter(|(id, _data)| {
                 let key_from_id = Pubkey::from((*id).clone().to_bytes());
-                if self.our_prices.contains(&key_from_id) {
-                    true
+                if let Some(market_hours) = self.our_prices.get(&key_from_id) {
+                    let ret = market_hours.can_publish_at(&now);
+
+		    if !ret {
+			debug!(self.logger, "Exporter: Attempted to publish price outside market hours";
```
ali-bahjati, line 488: The formatting is bad here again because of mixed tab and spaces

ali-bahjati (commented):
In general I recommend the flow to be from oracle to global store and from global store to the exporter to follow the design of agent.

src/agent/solana/exporter.rs
drozdziak1: Guilty as charged, I had a fairly old nightly locally while our CI grabs a fresh one with each run.

drozdziak1:
Re: proposed `oracle -> global store -> exporter` flow. 

I'm a bit on the fence about this. I think the proposed flow makes sense, but the existing permissions data flow is also a decent candidate, because of the similar "can I publish this?" concern. Perhaps the solution lies somewhere in the middle. Maybe both permissions and market hours could go through the global store.

jayantk (approved):

src/agent/market_hours.rs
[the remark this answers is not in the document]
jayantk: sorry the UTC comment was about line 46 above where you're using `Default::default()` to populate the `all_closed` object. I get that the `Tz` type parameter here makes timezone errors impossible (which is pretty cool actually)

src/agent/solana/exporter.rs
jayantk: there's a way to pin the version used for formatting which solves this problem. I forget exactly how to do it but we're doing it in pyth-crosschain.

PR #97, opened 2023-11-30
fix: use solana-client pubsub lib instead of solana-shadow

ali-bahjati:
Solana-shadow is not maintained anymore and doesn't work properly. It drops many updates and after a certain amount of latency drops all messages. We have been using Solana PubSub client in Hermes without any problem and there is no reason to stick with solana-shadow.

### DIFF
--- a/src/agent/solana/exporter.rs
+++ b/src/agent/solana/exporter.rs
@@ -481,25 +481,25 @@ impl Exporter {
                 if let Some(weekly_schedule) = self.our_prices.get(&key_from_id) {
                     let ret = weekly_schedule.can_publish_at(&now);
 
-		    if !ret {
-			debug!(self.logger, "Exporter: Attempted to publish price outside market hours";
-			       "price_account" => key_from_id.to_string(),
-			       "weekly_schedule" => format!("{:?}", weekly_schedule),
-			       "utc_time" => now.format("%c").to_string(),
-			       );
-		    }
+                    if !ret {
+                        debug!(self.logger, "Exporter: Attempted to publish price outside market hours";
+                            "price_account" => key_from_id.to_string(),
+                            "weekly_schedule" => format!("{:?}", weekly_schedule),
+                            "utc_time" => now.format("%c").to_string(),
+                        );
+                    }
 
-		    ret
+                    ret
                 } else {
                     // Note: This message is not an error. Some
                     // publishers have different permissions on
                     // primary/secondary networks
                     debug!(
-                    self.logger,
-                    "Exporter: Attempted to publish a price without permission, skipping";
-                    "unpermissioned_price_account" => key_from_id.to_string(),
-                    "permissioned_accounts" => format!("{:?}", self.our_prices)
-                            );
+                        self.logger,
+                        "Exporter: Attempted to publish a price without permission, skipping";
+                        "unpermissioned_price_account" => key_from_id.to_string(),
+                        "permissioned_accounts" => format!("{:?}", self.our_prices)
+                    );
                     false
                 }
             })
--- a/src/agent/solana/oracle.rs
+++ b/src/agent/solana/oracle.rs
@@ -632,19 +632,26 @@ mod subscriber {
             Result,
         },
         slog::Logger,
+        solana_account_decoder::UiAccountEncoding,
+        solana_client::{
+            nonblocking::pubsub_client::PubsubClient,
+            rpc_config::{
+                RpcAccountInfoConfig,
+                RpcProgramAccountsConfig,
+            },
+        },
         solana_sdk::{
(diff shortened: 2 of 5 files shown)


src/agent/solana/exporter.rs
```diff
@@ -481,25 +481,25 @@ impl Exporter {
...
                 } else {
                     // Note: This message is not an error. Some
                     // publishers have different permissions on
                     // primary/secondary networks
                     debug!(
-                    self.logger,
-                    "Exporter: Attempted to publish a price without permission, skipping";
-                    "unpermissioned_price_account" => key_from_id.to_string(),
-                    "permissioned_accounts" => format!("{:?}", self.our_prices)
-                            );
```
ali-bahjati, line 502: Formatting

src/agent/solana/oracle.rs
```diff
@@ -687,62 +694,70 @@ mod subscriber {
...
-            self.updates_tx
-                .send(shadow_rx.recv().await?)
+        pub async fn start(&self) -> Result<()> {
+            debug!(self.logger, "subscribed to program account updates"; "program_key" => self.program_key.to_string());
+
+            let client = PubsubClient::new(self.wss_url.as_str())
                 .await
-                .map_err(|_| anyhow!("failed to forward update"))
-        }
+                .expect("failed to create pubsub client");
```
ali-bahjati, line 724: Crashing the program on purpose here as currently in the design there is no systematic way to fail on startup errors.

src/agent/solana/oracle.rs
```diff
@@ -687,62 +694,70 @@ mod subscriber {
...
         }
 
-        async fn forward_update(
-            &self,
-            shadow_rx: &mut broadcast::Receiver<(Pubkey, Account)>,
-        ) -> Result<()> {
-            self.updates_tx
-                .send(shadow_rx.recv().await?)
+        pub async fn start(&self) -> Result<()> {
+            debug!(self.logger, "subscribed to program account updates"; "program_key" => self.program_key.to_string());
```
drozdziak1, line 720: Shouldn't this be logged at the end of pubsub client setup?

src/agent/solana/oracle.rs
```diff
@@ -687,62 +694,70 @@ mod subscriber {
...
+                .await?;
 
-            Ok(shadow.updates_channel())
+            loop {
+                match tokio_stream::StreamExt::next(&mut notif).await {
+                    Some(update) => {
+                        let account: Account = match update.value.account.decode() {
+                            Some(account) => account,
+                            None => {
+                                tracing::error!(?update, "Failed to decode account from update.");
```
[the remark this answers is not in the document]
ali-bahjati: right right. i missed this.

src/agent/solana/oracle.rs
```diff
@@ -687,62 +694,70 @@ mod subscriber {
...
-
-        async fn forward_updates(&self, shadow_rx: &mut broadcast::Receiver<(Pubkey, Account)>) {
             loop {
-                if let Err(err) = self.forward_update(shadow_rx).await {
+                let current_time = Instant::now();
+                if let Err(ref err) = self.start().await {
                     error!(self.logger, "{}", err);
                     debug!(self.logger, "error context"; "context" => format!("{:?}", err));
+                    if current_time.elapsed() < Duration::from_secs(30) {
+                        tracing::error!(
```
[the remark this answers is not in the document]
ali-bahjati: Fair. I used warn in the end because it means the rpc is not very reliable.

src/agent/solana/oracle.rs
ali-bahjati: It was there in the past but good point. Moved it below.

src/agent/solana/oracle.rs
ali-bahjati: My intention was to crash when the url is wrong in the beginning. but since we do this on reconnect what you say makes sense.

src/agent/solana/oracle.rs
```diff
@@ -674,75 +674,81 @@ mod subscriber {
...
-            Ok(shadow.updates_channel())
+            debug!(self.logger, "subscribed to program account updates"; "program_key" => self.program_key.to_string());
+
+            loop {
+                match tokio_stream::StreamExt::next(&mut notif).await {
+                    Some(update) => {
+                        let account: Account = match update.value.account.decode() {
+                            Some(account) => account,
+                            None => {
+                                tracing::error!(?update, "Failed to decode account from update.");
```
drozdziak1, line 736: This one too, please?

ali-bahjati: right.

drozdziak1 (approved):

PR #99, opened 2023-12-01
fix: use both latest publisher and aggregate slot for dynamic pricing

ali-bahjati:
One publisher has tried the previous version. I noticed that often they fall behind on some specific symbols by not publishing themself and that causes a sudden increase in priority once they publish it. That is why i am also considering the aggregate slot for the pricing. I'm also excluding the non-live feeds because publishers are often more unstable on them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2731,7 +2731,7 @@ dependencies = [
 
 [[package]]
 name = "pyth-agent"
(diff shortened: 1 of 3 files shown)


jayantk (approved):
👍  seems like a good adjustment. Now we only pay more if both the pub *and* the aggregate are behind.

PR #102, opened 2023-12-14
fix: fire and forget update tx

ali-behjati:
The old behaviour causes the exporter not to publish prices regularly. This change spawns a separate thread to unblock the exporter loop on the rpc send_tx call. Also, because of the batch staggering behaviour to avoid calling rpc for all batches at the same time, the old approach would always resulted in waiting longer than the publishing period.

### DIFF
--- a/src/agent/solana/exporter.rs
+++ b/src/agent/solana/exporter.rs
@@ -61,6 +61,7 @@ use {
             BTreeMap,
             HashMap,
         },
+        sync::Arc,
         time::Duration,
     },
     tokio::{
@@ -228,7 +229,7 @@ pub fn spawn_exporter(
 /// Exporter is responsible for exporting data held in the local store
 /// to the global Pyth Network.
 pub struct Exporter {
-    rpc_client: RpcClient,
+    rpc_client: Arc<RpcClient>,
 
     config: Config,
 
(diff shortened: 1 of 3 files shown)


src/agent/solana/exporter.rs
```diff
@@ -796,19 +803,28 @@ impl Exporter {
...
-                },
-            )
-            .await?;
-        debug!(self.logger, "sent upd_price transaction"; "signature" => signature.to_string(), "instructions" => instructions.len(), "price_accounts" => format!("{:?}", price_accounts));
-
-        self.inflight_transactions_tx.send(signature).await?;
+        let tx = self.inflight_transactions_tx.clone();
+        let logger = self.logger.clone();
+        let rpc_client =
+            RpcClient::new_with_timeout(self.rpc_url.clone().to_string(), self.rpc_timeout);
```
jayantk, line 809: I think it is slightly unfortunate that you are storing the url / timeout instead of the RpcClient above so that you can use `move` in the spawn call below. There is almost certainly a better way to do this -- you may be able to use`Arc`? David will know for sure.

jayantk (approved):

src/agent/solana/exporter.rs
ali-bahjati: Right. I thought the rpc request would be mutable for some reason but it is immutable and Arc works fine. Thanks for pointing it out.

PR #107, opened 2024-02-09
feat: wrap Pythnet/Solana accounts generically

Reisen:
See https://github.com/pyth-network/pyth-sdk-rs/pull/113 for how and why the API changed.

The short version is the API now provides the ability to read price accounts on both Solana and Pythnet accounting for the differences in those accounts. That API intentionally does not expose a generic wrapper because the differences (especially as new fields are added) cannot be abstracted.

In pyth-agent, the design currently does not distinguish between which network a price account came from in various code paths, but because these account structures now differ this will require a large refactor that introduces the concept of network differences. Rather than do this now, this introduces a thin wrapper around the API that _does_ act as a generic wrapper, which is possible because it currently doesn't care about extended fields. This is not a decent approach which is why it is not in the API, but is a good stopgap for pyth-agent as It's a small diff to unwind.

The wrapper works by copying the data into a fixed 64 sized account and ignoring the rest of the type entirely. It uses some hacks to do this unfortunately, but the alternative refactor for pyth-agent will be very large.

No new test is added to the integration tests because the tests use program admin to sync accounts and it doesn't work on newer oracle versions. We need to revamp the integration tests in the future to be in Rust and use new permissioning approaches.


src/agent/solana/oracle.rs
```diff
@@ -593,12 +661,13 @@ impl Poller {
             // as todo gets replaced with next_todo.
             for (price_key, price_account) in todo.iter().zip(price_accounts) {
                 if let Some(price_acc) = price_account {
-                    let price = load_price_account(&price_acc.data)
+                    let price = PriceEntry::load_from_account(&price_acc.data)
                         .context(format!("Could not parse price account at {}", price_key))?;
 
+                    let next_price = price.next;
```
guibescos, line 667: Do we need this extra variable?

src/agent/solana/oracle.rs
```diff
@@ -43,6 +49,66 @@ use {
...
+/// can iterate over either of these. The API is intended to force users to be aware of
+/// the account type they have, and so doesn't provide this abstraction (a good thing)
+/// and the agent should implement this in a better way.
+///
+/// For now, to implement the abstraction in the smallest way possible we use a shim
+/// type that uses the size of the accounts to determine the underlying representation
+/// and construct the right one regardless of which network we read. This will only work
+/// as long as we don't care about any extended fields.
+///
+/// TODO: Refactor the agent's network handling code.
```
jayantk, line 62: FWIW I think it's fine to keep things like this, because we're going to deprecate the solana oracle anyway, and then we won't need both account representations.

src/agent/solana/oracle.rs
Reisen: Yeah unfortunately the price gets moved above, rather than clone it this just copies out the only useful data we need.

src/agent/solana/oracle.rs
```diff
@@ -288,14 +353,17 @@ impl Oracle {
         account_key: &Pubkey,
         account: &Account,
     ) -> Result<()> {
-        let price_account = *load_price_account(&account.data)
+        // TODO: remove unwrap()
```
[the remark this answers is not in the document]
Reisen: Ah yeah thank you, removed comment.

src/agent/solana/oracle.rs
```diff
@@ -43,6 +49,66 @@ use {
...
+
+impl PriceEntry {
+    /// Construct the right underlying GenericPriceAccount based on the account size.
+    fn load_from_account(acc: &[u8]) -> Option<Self> {
+        unsafe {
+            let size = match acc.len() {
+                n if n == std::mem::size_of::<SolanaPriceAccount>() => 32,
+                n if n == std::mem::size_of::<PythnetPriceAccount>() => 64,
+                _ => return None,
+            };
```
guibescos, line 92: Do you really need unsafe? Why don't you conditionally deserialize as SolanaPriceAccount or PythnetPriceAccount.
Then you use the result SolanaPriceAccount or PythnetPriceAccount to build a PriceEntry.

Reisen: Can you expand on how to do that parsing? You might be right and I missed it but as far as I can tell these types are used mostly with zero copy casts and we haven't properly defined a parser/wire format to do this. In order to construct the underlying array I'm inferring from size, and the slice constructor is inherently unsafe which is where unsafe comes from.

Definitely would prefer a solution without unsafe, can you expand a bit more?

guibescos: Bytemuck try_from_bytes it like here : 
https://github.com/pyth-network/pyth-client/blob/f3b78939d768c525d29ed5fcebd767dc13e2fce8/program/rust/src/deserialize.rs#L33

guibescos: It's fine copying the data, you're copying it anyway

src/agent/pythd/adapter.rs
```diff
@@ -423,7 +426,7 @@ impl Adapter {
             publisher_accounts: price_account
                 .comp
                 .iter()
-                .filter(|comp| **comp != PriceComp::default())
+                .filter(|comp| *comp != &PriceComp::default())
```
ali-bahjati, line 429: Lack of this check is not harmful but having a default publisher in logs is annoying.

PR #109, opened 2024-02-23
fix: remove delay in publish_batch execution

ali-bahjati:
The existing code creates futures for publish_batch with delays to spread the rpc load evenly in the publish interval however due to the Rust async design those futures do not get started unless they are awaited. This results in additional deterministic latency of almost the publish_interval especially for the publishers who publish many prices.

This change removes the code as even the intended behaviour adds latency and disrupts the publish interval (the last publish_batch can block the actor loop).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3161,7 +3161,7 @@ dependencies = [
 
 [[package]]
 name = "pyth-agent"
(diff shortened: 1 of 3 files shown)


guibescos (approved):
lgtm

PR #111, opened 2024-03-28
chore: update dependencies

ali-bahjati:
This change updates the agent dependencies to their most recent version. Please see the inline comments on the code changes due to the upgrades for an explanation.

### DIFF
--- a/src/agent/dashboard.rs
+++ b/src/agent/dashboard.rs
@@ -15,7 +15,7 @@ use {
         },
     },
     crate::agent::metrics::MetricsServer,
-    chrono::NaiveDateTime,
+    chrono::DateTime,
     pyth_sdk::{
         Identifier,
         PriceIdentifier,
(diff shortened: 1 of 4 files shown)


src/agent/dashboard.rs
```diff
@@ -104,9 +102,7 @@ impl MetricsServer {
                 };
 
                 let last_local_update_string = if let Some(local_data) = price_data.local_data {
-                    if let Some(datetime) =
-                        NaiveDateTime::from_timestamp_opt(local_data.timestamp, 0)
-                    {
+                    if let Some(datetime) = DateTime::from_timestamp(local_data.timestamp, 0) {
```
ali-bahjati, line 105: This was a suggested change due to deprecation of the previous method, it returns a DateTime<Utc> which has the same behaviour as the NaiveDateTime for formatting (because the inner NaiveDateTime is the same and Utc offset is 0).

src/agent/dashboard.rs
```diff
@@ -176,9 +172,9 @@ pub struct DashboardSymbolView {
 
 #[derive(Debug)]
 pub struct DashboardPriceView {
-    local_data:      Option<PriceInfo>,
-    global_data:     Option<PriceEntry>,
-    global_metadata: Option<PriceAccountMetadata>,
+    local_data:       Option<PriceInfo>,
+    global_data:      Option<PriceEntry>,
+    _global_metadata: Option<PriceAccountMetadata>,
```
ali-bahjati, line 177: unused variable

src/bin/agent_migrate_config.rs
```diff
@@ -17,7 +17,7 @@ use {
     },
     toml_edit::{
         value,
-        Document,
+        DocumentMut,
```
ali-bahjati, line 20: Document was deprecated and DocumentMut was recommended as a workaround.

Reisen (approved):

PR #112, opened 2024-04-08
feat: add market_schedule module

keyvankhademi:
This add marker_schedule module and renames the old schedule to legacy_schedule.
It uses a new metadata key `schedule` and it is backward compatible by falling back to `weekly_schedule` metadata key.


src/agent/holiday_hours.rs
```diff
@@ -0,0 +1,376 @@
...
+            return Err(anyhow!("Invalid date format"));
+        }
+
+        let month: u32 = date_time_parts[0][..2]
+            .parse()
+            .map_err(|_| anyhow!("Invalid month"))?;
+        let day: u32 = date_time_parts[0][2..]
+            .parse()
+            .map_err(|_| anyhow!("Invalid day"))?;
+        let kind: HolidayDayKind = date_time_parts[1].parse()?;
```
Reisen, line 106: I don't think you are validating the day/month here? Should restrict these to day/month ranges. You also have a problem that it will depend on the month what day range is valid.

I think there is an assumption that the input string is trusted, if so can you comment that?

src/agent/holiday_hours.rs
```diff
@@ -0,0 +1,376 @@
...
+        Ok(())
+    }
+
+    #[test]
+    fn test_parsing_month_without_leading_zero_is_error() {
+        let s = "422/0900-1700";
+
+        let parsing_result: Result<HolidaySchedule> = s.parse();
+        assert!(parsing_result.is_err());
+    }
```
Reisen, line 258: I would suggest including the `proptest` and `arbitrary` crate's and adding a test that uses this to properly try and detect any parser failures, the test would look something like this after you define the arbitrary impls:

```
    proptest! {
        #[test]
        fn test_parse_schedule(holiday_schedule in holiday_schedule_strategy()) {
            let schedule_string = holiday_schedule.to_string();
            let parsed_schedule = schedule_string.parse::<HolidaySchedule>();
            prop_assert!(parsed_schedule.is_ok());
            prop_assert_eq!(parsed_schedule.unwrap(), holiday_schedule);
        }
    }
```

Reisen (changes requested):
I might be missing context, but is the string format for these holiday schedules defined anywhere or is it being introduced by this PR? I think it would be worse using a parsing library here to make the parsing a bit more robust. While `FromStr` is the right place to put this code I think handling parsing more robustly via something like `winnow` might be the right thing to do to define this format more concretely.

src/agent.rs
```diff
@@ -63,6 +63,7 @@ Note that there is an Oracle and Exporter for each network, but only one Local S
 ################################################################################################################################## */
 
 pub mod dashboard;
+pub mod holiday_hours;
```
ali-bahjati, line 66: I think we can group holidays and market hours to each other. market hours seems like a good name, one will be weekly schedule, and the other will be something like holiday overwrites

src/agent/holiday_hours.rs
```diff
@@ -0,0 +1,376 @@
...
+        Ok(HolidayDaySchedule { month, day, kind })
+    }
+}
+
+/// Helper enum for denoting per-day schedules: time range, all-day open and all-day closed.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum HolidayDayKind {
+    Open,
+    Closed,
+    TimeRange(NaiveTime, NaiveTime),
```
ali-bahjati, line 117: I think it would be nice to reuse HMKind in the other module and we eventually migrate them to the new format for consistency of our formats.

src/agent.rs
```diff
@@ -63,10 +63,10 @@ Note that there is an Oracle and Exporter for each network, but only one Local S
 ################################################################################################################################## */
 
 pub mod dashboard;
-pub mod market_hours;
 pub mod metrics;
 pub mod pythd;
 pub mod remote_keypair_loader;
+pub mod schedule;
```
keyvankhademi, line 69: module name: `market_schedule`

src/agent/schedule.rs
```diff
@@ -0,0 +1,14 @@
+pub mod holiday_hours;
+pub mod market_hours;
```
keyvankhademi, line 2: module name: `weekly_schedule`

src/agent/schedule.rs
```diff
@@ -0,0 +1,14 @@
+pub mod holiday_hours;
```
keyvankhademi, line 1: module name: `holiday_schedule`

src/agent/schedule.rs
keyvankhademi: rename `weekly_schedule` and `market_hours` to `legacy_schedule`

src/agent/schedule.rs
keyvankhademi: rename to `market_schedule`.

src/agent/holiday_hours.rs
```diff
@@ -0,0 +1,376 @@
...
+
+/// Holiday hours schedule
+#[derive(Clone, Default, Debug, Eq, PartialEq)]
+pub struct HolidaySchedule {
+    pub timezone: Option<Tz>,
+    pub days:     Vec<HolidayDaySchedule>,
+}
+
+impl HolidaySchedule {
+    pub fn all_closed() -> Self {
```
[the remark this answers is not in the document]
keyvankhademi: Fixed

src/agent/holiday_hours.rs
keyvankhademi: comment added.

src/agent.rs
keyvankhademi: Done

src/agent.rs
keyvankhademi: done

src/agent/schedule.rs
keyvankhademi: done

src/agent/schedule.rs
keyvankhademi: done

src/agent/holiday_hours.rs
keyvankhademi: not applicable anymore

src/agent/market_schedule.rs
```diff
@@ -0,0 +1,356 @@
...
+        PResult,
+        Parser,
+    },
+};
+
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct MarketSchedule {
+    pub timezone:        Tz,
+    pub weekly_schedule: Vec<ScheduleDayKind>,
```
keyvankhademi, line 43: maybe use array instead of vec because it has a fixed size.

src/agent/market_schedule.rs
```diff
@@ -0,0 +1,356 @@
...
+
+fn holiday_day_schedule_parser<'s>(input: &mut &'s str) -> PResult<HolidayDaySchedule> {
+    let ((month_str, day_str), kind) =
+        separated_pair((take(2usize), take(2usize)), "/", schedule_day_kind_parser)
+            .parse_next(input)?;
+
+    // day and month are not validated to be correct dates
+    // if they are invalid, it will be ignored since there
+    // are no real dates that match the invalid input
+    let month = month_str.parse::<u32>().unwrap();
```
keyvankhademi, line 115: avoid unwrap

src/agent/market_schedule.rs
```diff
@@ -0,0 +1,356 @@
...
+        }
+
+        self.weekly_schedule[weekday].can_publish_at(time)
+    }
+}
+
+fn market_schedule_parser<'s>(input: &mut &'s str) -> PResult<MarketSchedule> {
+    seq!(
+        MarketSchedule {
+            timezone: take_till(0.., ';').verify(|s| Tz::from_str(s).is_ok()).map(|s| Tz::from_str(s).unwrap()),
```
keyvankhademi, line 80: check how we can avoid parsing twice

src/agent/solana/oracle.rs
```diff
@@ -627,11 +629,45 @@ impl Poller {
...
+                        Some(schedule.unwrap())
+                    } else {
+                        warn!(
+                            self.logger,
+                            "Oracle: Product has schedule defined but it could not be parsed. Falling back to legacy schedule.";
+                            "product_key" => product_key.to_string(),
+                            "schedule" => msched_val,
+                        );
+                        None
+                    }
```
keyvankhademi, line 649: use match here

src/agent/solana/oracle.rs
```diff
@@ -627,11 +629,45 @@ impl Poller {
...
+                                ScheduleDayKind::from_mhkind(legacy_schedule.mon),
+                                ScheduleDayKind::from_mhkind(legacy_schedule.tue),
+                                ScheduleDayKind::from_mhkind(legacy_schedule.wed),
+                                ScheduleDayKind::from_mhkind(legacy_schedule.thu),
+                                ScheduleDayKind::from_mhkind(legacy_schedule.fri),
+                                ScheduleDayKind::from_mhkind(legacy_schedule.sat),
+                                ScheduleDayKind::from_mhkind(legacy_schedule.sun),
+                            ],
+                            holidays:        vec![],
+                        }),
```
keyvankhademi, line 670: use From trait here

src/agent/market_schedule.rs
```diff
@@ -0,0 +1,356 @@
...
+            .parse_next(&mut s.to_owned().as_str())
+            .map_err(|e| anyhow!(e))
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum ScheduleDayKind {
+    Open,
+    Closed,
+    TimeRange(NaiveTime, NaiveTime),
```
keyvankhademi, line 134: add support for 24:00

src/agent/market_schedule.rs
```diff
@@ -0,0 +1,356 @@
...
+        // Date match at end of range
+        assert!(market_schedule
+            .can_publish_at(&NaiveDateTime::parse_from_str("2023-04-22 17:00", format)?.and_utc()));
+
+        // Date match after range
+        assert!(!market_schedule
+            .can_publish_at(&NaiveDateTime::parse_from_str("2023-04-22 17:01", format)?.and_utc()));
+        Ok(())
+    }
+}
```
keyvankhademi, line 356: add good tests

src/agent/market_schedule.rs
keyvankhademi: decided not to use array because the parse works really well with vectors.

src/agent/market_schedule.rs
keyvankhademi: Fixed.

src/agent/solana/oracle.rs
keyvankhademi: Fixed.

src/agent/market_schedule.rs
keyvankhademi: Done

src/agent/market_schedule.rs
keyvankhademi: Done

src/agent/solana/oracle.rs
keyvankhademi: Fixed

src/agent/holiday_hours.rs
keyvankhademi: Done

src/agent/market_schedule.rs
keyvankhademi: Done

src/agent/market_schedule.rs
```diff
@@ -0,0 +1,559 @@
...
+    ) -> MarketSchedule {
+        MarketSchedule {
+            timezone: tz,
+            weekly_schedule,
+            holidays,
+        }
+    }
+}
+
+const VALID_SCHEDULE_DAY_KIND_REGEX: &str =
```
ali-bahjati, line 521: I recommend commenting what matches this regex.
It seems that it doesn't match hours > 12, is that intended?

ali-bahjati (approved):
Very nice!

src/agent/market_schedule.rs
keyvankhademi: Thats an oversight. I fixed it to support 24 hour format.

src/agent/legacy_schedule.rs
```diff
@@ -28,8 +28,10 @@ lazy_static! {
 }
 
 /// Weekly market hours schedule
+/// TODO: Remove after the migration
```
jayantk, line 31: nitpick: "after the migration" is not very specific. someone coming back to this todo won't necessarily know if it is done. Be precise: "after all publishers have upgraded to support the new schedule format"

jayantk (approved):

src/agent/legacy_schedule.rs
keyvankhademi: Makes sense, thanks. Fixed.

Reisen (approved):

src/agent/market_schedule.rs
```diff
@@ -0,0 +1,561 @@
...
+        let weekday = when_local.weekday().number_from_monday().to_usize();
+
+        for holiday in &self.holidays {
+            // Check if the day matches
+            if holiday.month == month && holiday.day == day {
+                return holiday.kind.can_publish_at(time);
+            }
+        }
+
+        self.weekly_schedule[weekday].can_publish_at(time)
```
jm-kr, line 108: last Sunday we had an index out of bounds error here, and our pyth-agent panicked.
```shell
thread 'tokio-runtime-worker'
panicked at
index out of bounds:
src/agent/market_schedule.rs:108:29:
the len is 7 but the index is 7
note: run with 'RUST_BACKTRACE=1'
environment variable to display a backtrace
```

keyvankhademi: Thanks for catching this. I pushed the fixed in another PR.

PR #114, opened 2024-04-16
fix: weekday calculation bug

keyvankhademi:


src/agent/market_schedule.rs
```diff
@@ -96,7 +96,7 @@ impl MarketSchedule {
         let month = when_local.date_naive().month0() + 1;
         let day = when_local.date_naive().day0() + 1;
         let time = when_local.time();
-        let weekday = when_local.weekday().number_from_monday().to_usize();
+        let weekday0 = when_local.weekday().number_from_monday().to_usize() - 1;
```
ali-bahjati, line 99: I recommend adding a test for Monday too. `0-1` might overflow.

ali-bahjati (approved):

src/agent/market_schedule.rs
keyvankhademi: Done

PR #117, opened 2024-04-22
refactor: migrate adapter service to api

Reisen:
This PR refactors the Adapter to remove the running service component. The goal behind this is to remove the channel communication between the adapter and other components, instead, Adapter now acts as an API. The Adapter API currently is a single nameless trait with all functionality combined, this change will allow splitting that API out into relevant components to better isolate functionality. For now though, this just moves the entire API into a temporary trait called `AdapterApi`, in future PR's, `adapter/` will become `state/` and the relevant functionality will be broken out.

Apart from the `AdapterApi` and channel removal, all other changes are mechanical, just propagating the change through the codebase.

Some changes that occurred during this refactoring:

- Tests from rpc.rs are removed, as they don't actually test anything meaningful (just whether a send/recv over a channel with hardcoded values works)
- A temporary notifier service is in `.../adapter/api.rs` to replace the only meaningful runtime component of adapter.
- An abstract `S` parameter is added anywhere Adapter used to be relied on.
- The counter logic mildly changed as it is now an atomic, which makes the `++increment` behaviour `increment++`
- The `AdapterApi` is now strongly typed, so the rpc.rs API now does proper verification of accounts in the right place.
- Non-trait methods factored out of `AdapterApi`

### DIFF
--- a/src/agent/pythd/api/rpc.rs
+++ b/src/agent/pythd/api/rpc.rs
@@ -48,6 +48,7 @@ use {
     std::{
         fmt::Debug,
         net::SocketAddr,
+        sync::Arc,
     },
     tokio::sync::{
         broadcast,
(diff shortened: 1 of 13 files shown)


src/agent/pythd/api/rpc.rs
```diff
@@ -477,792 +499,3 @@ async fn serve(
 
     tokio::task::spawn(serve).await.map_err(|e| e.into())
 }
-
-#[cfg(test)]
```
Reisen, line 481: Actually there is some value to these, I will bring these back but in a slightly different form.

ali-bahjati (approved):

src/agent/pythd/api/rpc.rs
ali-bahjati: i think the adaptor tests are a bit useless but this is good.

PR #121, opened 2024-05-10
feat: add publish_interval logic

keyvankhademi:
This PR introduces support for a new product metadata key `publish_interval`. It will only allow the publish of a new price after this interval has passed since last updated price.

### DIFF
--- a/src/agent/solana/exporter.rs
+++ b/src/agent/solana/exporter.rs
@@ -497,22 +497,24 @@ impl Exporter {
             })
             .filter(|(id, info)| {
                 // Filtering out prices that are being updated too frequently according to publisher_permission.publish_interval
-                let last_info = self.last_published_state.get(id);
-                if last_info.is_none() {
-                    // No prior data found, letting the price through
-                    return true;
-                }
-                let last_info = last_info.unwrap();
+                let last_info = match self.last_published_state.get(id) {
+                    Some(last_info) => last_info,
+                    None => {
+                        // No prior data found, letting the price through
+                        return true;
+                    }
+                };
 
                 let key_from_id = Pubkey::from((*id).clone().to_bytes());
-                let publisher_permisssion = self.our_prices.get(&key_from_id);
-                if publisher_permisssion.is_none() {
-                    // Should never happen since we have filtered out the price above
-                    return false;
-                }
-                let publisher_permission = publisher_permisssion.unwrap();
+                let publisher_metadata = match self.our_prices.get(&key_from_id) {
+                    Some(metadata) => metadata,
+                    None => {
+                        // Should never happen since we have filtered out the price above
+                        return false;
+                    }
+                };
 
-                if let Some(publish_interval) = publisher_permission.publish_interval {
+                if let Some(publish_interval) = publisher_metadata.publish_interval {
                     if info.timestamp < last_info.timestamp + publish_interval {
                         // Updating the price too soon after the last update, skipping
                         return false;
--- a/src/agent/solana/exporter.rs
+++ b/src/agent/solana/exporter.rs
@@ -8,6 +8,7 @@ use {
         },
         key_store,
         network::Network,
+        oracle::PublisherPermission,
     },
     crate::agent::{
         market_schedule::MarketSchedule,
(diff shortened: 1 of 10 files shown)


src/agent/solana/exporter.rs
```diff
@@ -501,6 +494,31 @@ impl Exporter {
...
                 }
             })
+            .filter(|(id, info)| {
+                // Filtering out prices that are being updated too frequently according to publisher_permission.publish_interval
+                let last_info = self.last_published_state.get(id);
+                if last_info.is_none() {
+                    // No prior data found, letting the price through
+                    return true;
+                }
+                let last_info = last_info.unwrap();
```
ali-bahjati, line 504: `match last_info { ... }` is cleaner here.

src/agent/solana/exporter.rs
```diff
@@ -501,6 +494,31 @@ impl Exporter {
...
+
+                let key_from_id = Pubkey::from((*id).clone().to_bytes());
+                let publisher_permisssion = self.our_prices.get(&key_from_id);
+                if publisher_permisssion.is_none() {
+                    // Should never happen since we have filtered out the price above
+                    return false;
+                }
+                let publisher_permission = publisher_permisssion.unwrap();
```
ali-bahjati, line 512: same here.

src/agent/solana/exporter.rs
```diff
@@ -501,6 +494,31 @@ impl Exporter {
...
+
+                if let Some(publish_interval) = publisher_permission.publish_interval {
+                    if info.timestamp.saturating_sub(last_info.timestamp) < publish_interval.as_secs() as i64 {
```
ali-bahjati, line 515: you are falling back to seconds here too. I recommend changing the info.timestamp and now to `Instant` ([link](https://docs.rs/tokio/latest/tokio/time/struct.Instant.html)) so you get duration_since and can compare properly.

src/agent/solana/exporter.rs
keyvankhademi: We have 3 `Options` here that we have to unwrap. I initially did the `match` and `let Some(...)` but it looked really ungly and unreadable. I think this way atleast it avoids 3 layers of nested matches.

src/agent/solana/exporter.rs
keyvankhademi: So we have a problem here. we are only storing seconds for `last_info.timestamp`. To fix this to work with values <1s we have to change the info logic to store NaiveTime instead of u64. I'll look into that to see how much effort its gonna take.

ali-bahjati: Why not the Instant that I mentioned?

keyvankhademi: Can you explain more how you want to measure the exact time with that? Because we have an exact duration we want to enforce. The Instant create mentions that

> Note, however, that instants are not guaranteed to be steady. In other words, each tick of the underlying clock may not be the same length (e.g. some seconds may be longer than others). An instant may jump forwards or experience time dilation (slow down or speed up), but it will never go backwards.

It seems like it's not for the purpose of calculating exact times. Am I missing something?

src/agent/solana/exporter.rs
ali-bahjati: You can make it:
```rust
let last_info = match self.last_published_state.get(id) {
  Some(info) => info,
  None => {
    return true,
  }
}
```

ali-bahjati (approved):
Nice! Please address my inline comment before merging.

src/agent/solana/exporter.rs
keyvankhademi: Perfect. Thanks.

src/agent/solana/exporter.rs
keyvankhademi: Done

src/agent/solana/exporter.rs
keyvankhademi: I tried using `tokio::time::Instant` and `std::time::Instant`. It could have potentially worked since we are only comparing some `Instant::now()` times. But they don't allow access to create or read the underlying time data. In `metrics.rs` we are storing the exact time for the prices and there is no way to convert Instant to a unix timestamp.
I eventually went with NaiveDate .

PR #122, opened 2024-05-14
refactor(agent): convert global store to an Api

Reisen:
This PR follows on from previous refactoring to change the global store from a running service to an API over the recently introduced `Adapter` refactor. While doing so, the `global.rs` module was moved and the code changed to reflect the new function calls that replace the previous channel based communication.

This PR has a followup PR with the same changes to the `local.rs` module. This PR is part of the following refactor:

- #116 
- #117


src/agent.rs
```diff
@@ -118,26 +119,29 @@ impl Agent {
...
-        let (secondary_oracle_updates_tx, secondary_oracle_updates_rx) =
-            mpsc::channel(self.config.channel_capacities.secondary_oracle_updates);
-        let (global_store_lookup_tx, global_store_lookup_rx) =
-            mpsc::channel(self.config.channel_capacities.global_store_lookup);
         let (local_store_tx, local_store_rx) =
             mpsc::channel(self.config.channel_capacities.local_store);
         let (primary_keypair_loader_tx, primary_keypair_loader_rx) = mpsc::channel(10);
         let (secondary_keypair_loader_tx, secondary_keypair_loader_rx) = mpsc::channel(10);
 
+        let mut registry = PROMETHEUS_REGISTRY.lock().await;
```
keyvankhademi, line 127: This feels out of place. If we are just passing it to the adapter then it's passed down to global store, can't we initialize it in the global store?

src/agent/pythd/adapter/global.rs
```diff
@@ -0,0 +1,352 @@
...
+    async fn accounts_metadata(&self) -> Result<AllAccountsMetadata>;
+    async fn accounts_data(&self, network: Network) -> Result<AllAccountsData>;
+    async fn price_accounts(
+        &self,
+        network: Network,
+        price_ids: HashSet<Pubkey>,
+    ) -> Result<HashMap<Pubkey, PriceEntry>>;
+}
+
+// Allwo downcasting Adapter into GlobalStore for functions that depend on the `GlobalStore` service.
```
keyvankhademi, line 179: typo `Allow`

src/agent.rs
Reisen: Yeah this is temporary. Before this change the static global one was used but my changes caused initialization code to not be async (good) but therefore no longer able to take the lock in those places (bad). This is also used in other places so as I refactor the other parts of the code, they will also need the metrics. So lifting it up to the top so I can pass it through temporarily is what I opted for.

Later, I will do what we do in hermes, and have a `Metrics` API that does the initialization and the other APIs will just use it exactly as you suggest.

src/agent/pythd/adapter/global.rs
```diff
@@ -0,0 +1,352 @@
...
+                    .get(&id)
+                    .cloned()
+                    .map(|v| (id, v))
+                    .ok_or(anyhow!("price id not found"))
+            })
+            .collect()
+    }
+}
+
+async fn update_data<S>(state: &S, network: Network, update: &Update) -> Result<()>
```
ali-bahjati, line 236: any reason you couldn't make it part of the trait?

src/agent/pythd/adapter/global.rs
```diff
@@ -0,0 +1,352 @@
...
+pub enum Lookup {
+    LookupAllAccountsMetadata,
+    LookupAllAccountsData {
+        network: Network,
+    },
+    LookupPriceAccounts {
+        network:   Network,
+        price_ids: HashSet<Pubkey>,
+    },
+}
```
ali-bahjati, line 113: Do we need this?

src/agent/pythd/adapter/global.rs
ali-bahjati: ahaa it's actually a private function. alright.

ali-bahjati (approved):
It seems very nice to me. I love seeing code getting simpler.

I don't love Adaptor name and it's position in the code. I think in the next PR we should change the name of the Adapter to a better name (like State) and move it out of pythd.

p.s: i think in the following PRs might be good to do mechanical moves as a separate commit so reviewing changes be easier. Github diff and difftastic couldn't put removed and added global.rs es together. I know it's a lot to ask while doing this kinds of refactors. So not important that much.

ali-bahjati (approved):

PR #123, opened 2024-05-27
refactor(agent): convert local store to an Api

Reisen:
This PR follows on from previous refactoring to change the local store from a running service to an API over the recently introduced `Adapter` refactor.

This PR is part of the following refactor:

- #116 
- #117 
- #122


src/agent/pythd/adapter/local.rs
```diff
@@ -0,0 +1,110 @@
...
+            conf,
+            timestamp: _,
+        } = self;
+
+        status == &other.status && price == &other.price && conf == &other.conf
+    }
+}
+
+pub struct Store {
+    prices:  RwLock<HashMap<PriceIdentifier, PriceInfo>>,
```
ali-bahjati, line 49: not for this one as it's restructing, but in the future DashMap would make it better because update seems to be for each price feed.

ali-bahjati (approved):
Very nice!

PR #126, opened 2024-06-05
refactor(agent): various refactors

Reisen:
This PR contains several refactors, each commit is self-contained and should be reviewed individually. You can leave comments directly on the commits. See the commit messages which might contain more information about a specific refactor.

Follow-up to:

- #116 
- #117 
- #124 
- #125


src/agent/state/global.rs
```diff
@@ -279,23 +275,6 @@ where
...
-                    Identifier::new(account_key.to_bytes()),
-                    account.agg.price,
-                    account.agg.conf,
-                    account.agg.status,
-                    account.valid_slot,
-                    account.agg.pub_slot,
-                )
-                .await
-                .map_err(|_| anyhow!("failed to notify pythd adapter of account update"))?;
-            }
```
ali-bahjati, line 298: Why is this one removed?

ali-bahjati (commented):
Very nice! I think one function call is removed by mistake but LGTM apart from that.

src/agent/state/keypairs.rs
```diff
@@ -0,0 +1,281 @@
...
+
+    async fn update_keypair(&self, network: Network, new_keypair: Keypair) {
+        *match network {
+            Network::Primary => self.into().primary_current_keypair.write().await,
+            Network::Secondary => self.into().secondary_current_keypair.write().await,
+        } = Some(new_keypair);
+    }
+}
+
+pub async fn spawn<S>(
```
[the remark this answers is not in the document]
Reisen: Yeah I did want to move this out, but wanted to finish the overall refactors before doing so, I can do it in a future PR with some further refactors.

src/agent.rs
```diff
@@ -78,11 +70,33 @@ use {
...
+    /// gracefully shut down the application.
+    ///
+    /// We make this global based on the fact the:
+    /// - The `Sender` side does not rely on any async runtime.
+    /// - Exit logic doesn't really require carefully threading this value through the app.
+    /// - The `Receiver` side of a watch channel performs the detection based on if the change
+    ///   happened after the subscribe, so it means all listeners should always be notified
+    ///   correctly.
+    pub static ref EXIT: watch::Sender<bool> = watch::channel(false).0;
+}
```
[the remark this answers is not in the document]
Reisen: Right, the original code also did not, this is just converting to the static to match hermes, I didn't want to touch exit behaviour just yet.

src/agent/state/global.rs
Reisen: It's moved into the Prices service, otherwise you have Prices -> GlobalStore -> Prices, I left a comment in the commit but just to expand further. The only places this is called is from oracle. I renamed the existing `update_price` in `Prices` to `update_local_price`, and as the logic for updating is already there, it made more sense to also add `update_global_price` as well, and then moving this logic into Pricing, given the function is already defined there.

I'm pretty confident this is a better design and how it really should have been, especially given the existence of `update_price`.

src/agent/state.rs
```diff
@@ -1392,60 +1368,106 @@ mod tests {
...
+                Network::Primary,
+                &Update::PriceAccountUpdate {
+                    account_key: account,
+                    account:     test_price.into(),
+                }, // Identifier::new(account.to_bytes()),
+                   // price,
+                   // conf,
+                   // PriceStatus::Trading,
+                   // valid_slot,
+                   // pub_slot,
```
ali-bahjati, line 1449: missed comment

src/agent/state/keypairs.rs
```diff
@@ -0,0 +1,282 @@
...
+    vec![http_api_jh]
+}
+
+/// Validate and apply a keypair to the specified mut reference,
+/// hiding errors in logs.
+///
+/// Returns the appropriate HTTP response depending on checks success.
+///
+/// NOTE(2023-03-22): Lifetime bounds are currently necessary
+/// because of https://github.com/rust-lang/rust/issues/63033
```
keyvankhademi, line 210: I think this issue is now resolved, we might need to update rust tho. Maybe we can fix this in a future refactor PR.

src/lib.rs
```diff
@@ -1,8 +1 @@
-// The typed-html crate does pretty deep macro calls. Bump if
-// recursion limit compilation errors return for html!() calls.
-#![recursion_limit = "256"]
```
keyvankhademi, line 3: Are you sure we don't need this?

ali-bahjati: I think he removed the dashboard entirely in previous prs

keyvankhademi: Then I guess we should remove the package from cargo if it's not used anymore.

PR #128, opened 2024-06-20
refactor(agent): extract oracle component/service

Reisen:


src/agent/config.rs
```diff
@@ -0,0 +1,85 @@
...
+            .add_source(Environment::with_prefix("agent"))
+            .build()?
+            .try_deserialize()
+            .map_err(|e| e.into())
+    }
+}
+
+/// Capacities of the channels top-level components use to communicate
+#[derive(Deserialize, Debug)]
+pub struct ChannelCapacities {
```
ali-bahjati, line 53: might be good to mark the deprecated ones.

src/agent/services/oracle.rs
```diff
@@ -0,0 +1,206 @@
...
+            config.rpc_timeout,
+            CommitmentConfig {
+                commitment: config.oracle.commitment,
+            },
+        ))
+    };
+
+    loop {
+        select! {
+            update = notifier.next()      => handle_account_update(state.clone(), network, update).await,
```
ali-bahjati, line 114: how we handle websocket failures here? i don't see any logic to handle it. i think we had it in the past.

src/agent/services/oracle.rs
```diff
@@ -0,0 +1,206 @@
...
+                None => {
+                    tracing::error!(
+                        update = ?update,
+                        "Failed to decode account from update.",
+                    );
+                }
+            };
+        }
+        None => {
+            tracing::debug!("subscriber closed connection");
```
ali-bahjati, line 167: back to my comment above, we need to reconnect when it happens.

src/agent/solana.rs
```diff
@@ -70,25 +76,51 @@ pub mod network {
         pub exporter:    exporter::Config,
     }
 
+    /// Spawn an Oracle, in-progress porting this to State.
+    ///
+    /// Behaviour:
+    /// - Spawns Oracle: (Obsolete, now Extracted to state/oracle.rs)
+    ///   - Spawns a Subscriber:
```
ali-bahjati, line 83: subscriber is no more

src/agent/state/oracle.rs
```diff
@@ -0,0 +1,618 @@
...
+            status = ?price_entry.agg.status,
+            "Observed on-chain price account update.",
+        );
+
+        self.into()
+            .data
+            .write()
+            .await
+            .price_accounts
+            .insert(*account_key, price_entry.clone());
```
ali-bahjati, line 252: i recall that there were some logics to make sure we only store the latest price, maybe you removed it because it's always more recent?

ali-bahjati (commented):
Nice! The code is now much simpler!! thanks! I think we want to fix the WS behaviour and CI before merging.

src/agent/state/oracle.rs
ali-bahjati: my bad it's somewhere else. sorry.

PR #130, opened 2024-07-01
refactor(agent): add tracing to expensive calls

Reisen:
Adds some initial tracing to the new refactored version of the agent based on API method calls. The update_price RPC call emits a trace log with an account key, and the batch function is similarly emitting keys, so it should now be possible  to measure the time difference between an update entering the program and being submitted.


ali-bahjati:
This changeset was used in #131

PR #131, opened 2024-07-12
improve tracing coverage

cctdaniel:
This PR enhances tracing by introducing the `#[instrument]` macro to more functions and adding additional fields to the current span in `get_product.rs`, `subscribe_price.rs`, and `subscribe_price_sched.rs`. This approach is async safe as it maintains the reference to the current span without manual enter/exit, ensuring the tracing context is preserved across await points. We should avoid manual enter/exit tracing logs already handled by `#[instrument]`, and any custom spans should be cautious of this behavior to avoid manual span entry.

### DIFF
--- a/config/config.sample.pythtest.toml
+++ b/config/config.sample.pythtest.toml
@@ -60,3 +60,12 @@ exporter.compute_unit_price_micro_lamports = 1000
 # Note that this doesn't affect the rate at which transactions are published:
 # this is soley a backwards-compatibility API feature.
 notify_price_sched_interval_duration = "400ms"
+
+# Configuration for OpenTelemetry
+[opentelemetry]
+
+# Timeout in seconds for the OpenTelemetry exporter
+exporter_timeout_secs = 3
+
+# Endpoint URL for the OpenTelemetry exporter
+exporter_endpoint = "http://127.0.0.1:4317"
--- a/config/config.sample.pythtest.toml
+++ b/config/config.sample.pythtest.toml
@@ -60,12 +60,3 @@ exporter.compute_unit_price_micro_lamports = 1000
 # Note that this doesn't affect the rate at which transactions are published:
 # this is soley a backwards-compatibility API feature.
 notify_price_sched_interval_duration = "400ms"
-
-# Configuration for OpenTelemetry
-[opentelemetry]
-
-# Timeout in seconds for the OpenTelemetry exporter
-exporter_timeout_secs = 3
-
-# Endpoint URL for the OpenTelemetry exporter
-exporter_endpoint = "http://127.0.0.1:4317"
--- a/src/agent/config.rs
+++ b/src/agent/config.rs
@@ -32,6 +32,8 @@ pub struct Config {
     pub metrics_server:        metrics::Config,
     #[serde(default)]
     pub remote_keypair_loader: services::keypairs::Config,
+    #[serde(default)]
+    pub opentelemetry:         OpenTelemetryConfig,
 }
 
 impl Config {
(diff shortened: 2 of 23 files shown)


src/agent/pyth/rpc/get_product.rs
```diff
@@ -27,6 +29,8 @@ where
     }?;
 
     let account = params.account.parse::<solana_sdk::pubkey::Pubkey>()?;
+    tracing::Span::current().record("account", account.to_string());
```
Reisen, line 32: Nitpick: This is the right way to add additional fields in the current span, would you mind modifying the PR description to point out that these tracing changes modify the current span introduced by `instrument` and is therefore async safe because it handles maintaining the ref (see: https://docs.rs/tracing/latest/tracing/struct.Span.html#method.enter) - we should make sure to avoid tracing logs that are handled by instrument and/or avoid entering new spans manually as you did here so I think as this is the first PR doing this it's worth documenting.

Reisen (approved):

src/bin/agent.rs
```diff
@@ -31,18 +35,38 @@ struct Arguments {
 #[tokio::main]
 async fn main() -> Result<()> {
     // Initialize a Tracing Subscriber
-    let fmt_builder = tracing_subscriber::fmt()
+    let fmt_layer = tracing_subscriber::fmt::layer()
         .with_file(false)
         .with_line_number(true)
         .with_thread_ids(true)
-        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .with_target(true)
```
ali-bahjati, line 42: why is this :?

cctdaniel: ah good catch, seems like default is true, so we can remove this line https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/src/fmt/format/mod.rs#L600

src/bin/agent.rs
```diff
@@ -31,18 +35,38 @@ struct Arguments {
...
         .with_ansi(std::io::stderr().is_terminal());
 
+    // Set up the OpenTelemetry exporter, defaults to 127.0.0.1:4317
```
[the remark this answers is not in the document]
cctdaniel: good point

config/config.sample.pythtest.toml
```diff
@@ -60,3 +60,12 @@ exporter.compute_unit_price_micro_lamports = 1000
 # Note that this doesn't affect the rate at which transactions are published:
```
ali-bahjati, line 60: let's hide it from here for the time being.

src/agent/config.rs
```diff
@@ -32,6 +32,8 @@ pub struct Config {
     pub metrics_server:        metrics::Config,
     #[serde(default)]
     pub remote_keypair_loader: services::keypairs::Config,
+    #[serde(default)]
+    pub opentelemetry:         OpenTelemetryConfig,
```
ali-bahjati, line 36: Let's make this one an option and actually not construct it if it's not given. I know that it's probably harmless but it's better to be clear.

src/bin/agent.rs
```diff
@@ -30,21 +34,6 @@ struct Arguments {
 
 #[tokio::main]
 async fn main() -> Result<()> {
-    // Initialize a Tracing Subscriber
-    let fmt_builder = tracing_subscriber::fmt()
-        .with_file(false)
-        .with_line_number(true)
-        .with_thread_ids(true)
-        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
```
ali-bahjati, line 38: I see this one is removed. What does this removal entail?

cctdaniel: ah this is the ability to allow tracing based on `RUST_LOG` env var, will add back the feature

config/config.sample.pythtest.toml
cctdaniel: im assuming you mean the opentelemetry config?

ali-bahjati (approved):
Nice! Can you also bump the version by minor?

PR #132, opened 2024-07-23
fix: use env filter properly

ali-bahjati:
Apparently if we pass env filter in an array with the other layers it won't be executed first (to act as a filter). This change fixes that by explicitly adding it before.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3400,7 +3400,7 @@ dependencies = [
 
 [[package]]
 name = "pyth-agent"
-version = "2.10.0"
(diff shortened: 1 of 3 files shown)


factec-kyle:
Even though I set RUST_LOG=info, the agent was still dumping logs including TRACE. This might have been the cause.

ali-bahjati:
@factec-kyle yeah that's right, this PR aims to fix that.

cctdaniel (approved):
🙇🏻‍♂️

PR #134, opened 2024-07-29
fix: spawn transaction senders in exporter

Reisen:
During refactoring the async spawning of the transaction sender was removed but not added back.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "stable"
+channel = "1.79.0"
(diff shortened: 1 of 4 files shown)


rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "stable"
+channel = "1.79.0"
```
ali-bahjati, line 2: thank you!

ali-bahjati (approved):
Nice! can you bump the version before merging?

PR #135, opened 2024-08-01
fix: feed rpc responses and notifications instead of send and flush

ali-bahjati:
We might send thousands of messages per second to the clients and `send` method flushes each request which might take a longer time. This change changes our responses and notifications to `feed` and flushes the messages based on a configured interval (default: 100ms).

There is a small refactor of a config in the previous changes to make it match our configuration style.

### DIFF
--- a/config/config.toml
+++ b/config/config.toml
@@ -6,6 +6,18 @@
 # connection is not exposed for unauthorized access.
 listen_address = "127.0.0.1:8910"
 
+# Size of the buffer of each Server's channel on which `notify_price` events are
+# received from the Price state.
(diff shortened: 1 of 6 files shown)


config/config.toml
```diff
@@ -187,7 +199,7 @@ key_store.mapping_key = "RelevantOracleMappingAddress"
 [opentelemetry]
 
 # Timeout in seconds for the OpenTelemetry exporter
```
Reisen, line 201: Should probably change this comment from "seconds" now that it can accept arbitrary units.

Reisen (approved):

PR #140, opened 2024-09-02
feat: support batch publish

Riateche:
- Add optional `publish_program_key` field to the config
- Fetch feed_id for each permissioned feed
- Create instruction for the publish program with fallback to the oracle program


src/agent/state/exporter.rs
```diff
@@ -534,46 +551,55 @@ where
...
+        updates.push(update);
+    }
 
+    if let Some(publish_program_key) = publish_program_key {
+        let (instruction, unsupported_updates) = create_instruction_with_publish_program(
+            publish_keypair.pubkey(),
+            publish_program_key,
+            updates,
+        )?;
+        updates = unsupported_updates;
```
ali-bahjati, line 584: I think the fallback might fail because we also change the batch size, so maybe a simpler way can be just to crash? I think it highly depends on our upgrade process.

Cargo.toml
```diff
@@ -56,6 +56,8 @@ tracing-opentelemetry = "0.24.0"
 opentelemetry = "0.23.0"
 opentelemetry_sdk = { version = "0.23.0", features = ["rt-tokio"]}
 opentelemetry-otlp = { version = "0.16.0" }
+pyth-price-publisher = { git = "https://github.com/pyth-network/pyth-crosschain", branch = "add-publisher-program" }
```
ali-bahjati, line 59: let's publish this package too.

ali-bahjati (approved):
Very nice!

src/agent/state/exporter.rs
```diff
@@ -771,9 +797,55 @@ fn create_instruction_without_accumulator(
...
+        } else {
+            values.push(PublisherPrice::new(
+                update.feed_index,
+                (update.info.status as u8).into(),
+                update.info.price,
+                update.info.conf,
+            )?);
+        }
+    }
+    let mut data = vec![1];
```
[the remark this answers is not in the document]
Riateche: Now it imports instruction types from the program crate.

ali-bahjati (approved):

Cargo.toml
Riateche: Done.

PR #145, opened 2024-09-13
refactor: remove mapping account usage

ali-bahjati:
Solana get program accounts RPC call is a great way to track price and product accounts instead of mapping account. Mapping account is not efficient and gets in the way if we change its structure. This PR removes all the usages of the mapping account.

Please review commit-by-commit.


Riateche (approved):
Nice!

PR #147, opened 2024-10-29
Fix: Update config samples with more descriptive pyth_oracle_ for

DaveWK:
program key

This change was made in the codebase already but didn't make it into the samples. Now that there's 2 program_key entries this makes it clear what they are used for


merolish:
Do we still want to make this change?

ali-bahjati (approved):

PR #148, opened 2025-01-02
Publish a Minimal Container Image to ECR Public

thebeanogamer:
Running the current full-size image in a production deployment isn't really feasible, but the minimal container image already in this repo already has everything required. We're looking to move our Pyth Publisher into Kubernetes and this would make that change much easier.

To give you some idea of the size difference:
```bash
$ podman images --format "table {{.ID}} {{.Repository}} {{.Size}}" | grep -E "(pyth-network|6a18b7d)"
6a18b7d60f62  <none>                             136 MB
774cbbe966e0  public.ecr.aws/pyth-network/agent  5.83 GB
```

It'd probably also be worthwhile to upload a basic `About` and Icon to ECR Public to help users know they've found the correct page. Some guidance on doing that can be found on <https://docs.aws.amazon.com/AmazonECR/latest/public/public-repository-catalog-data.html>.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 target
 Dockerfile
+agent.Dockerfile
--- a/.github/workflows/image-push.yaml
+++ b/.github/workflows/image-push.yaml
@@ -15,6 +15,13 @@ jobs:
     permissions:
       id-token: write
       contents: read
+    strategy:
+      matrix:
+        dockerfile:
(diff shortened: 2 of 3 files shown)


ali-bahjati:
Thanks a lot for your contribution.

I think the existing one is only useful for testing, i created #149 to actually replace the main one with the minimal one and bumped the version.

PR #151, opened 2025-02-27
Fix tokio task leak

rodoufu:
While running the agent  version `v.2.12.1`, it crashed after some time with an out of memory issue, having to be killed by the OS.
The machine I was using has 48GB of RAM

It was happening for both mainet and testnet.

I did not see anything particularly special about the resources it was using:
tokio_console_resources-pr

But the number of tasks seems large.
tokio_console_threads-pr

One can see 7721 tasks while a lot of them are idle for some time.
After observing it, the number of tasks would increase with time until the point it would, the binary would be killed.
The idle tasks are created here https://github.com/pyth-network/pyth-agent/blob/main/src/agent/services/oracle.rs#L132 as can be seen in the image, where the subscriber is handling the `handle_price_account_update`.
That line is crating tokio tasks without keeping track of the task handle `JoinHandle`, in cases where those are created faster than they are `.await`ed this can lead to a leak.

After the proposed change I can see a much more comfortable number of tasks:
tokio-console_threads_after-pr
The number of tasks now is stable around 100, being 114 in the attached image.
It is worthy mentioning that I used 100 worker tasks to wait for the previously leaked ones to finish, so this number can be way smaller with another configuration.

In order to reproduce the tokio console one can follow the instructions in https://github.com/tokio-rs/console which are basic:
- Install tokio-console with `cargo install --locked tokio-console`
- Add `console-subscriber = "0.3.0"` as a dependency
- Add `console_subscriber::init();` as the first line in the `main` funciton
- Run the binary with `RUSTFLAGS="--cfg tokio_unstable" cargo run --bin agent -- --config <config file path>`
- Run `tokio-console` to watch its data

### DIFF
--- a/src/agent/services/oracle.rs
+++ b/src/agent/services/oracle.rs
@@ -67,11 +67,11 @@ where
     )));
 
     if config.oracle.subscriber_enabled {
-        let number_of_workers = 100;
-        let channel_size = 1000;
-        let (sender, receiver) = tokio::sync::mpsc::channel(channel_size);
-        let max_elapsed_time = Duration::from_secs(30);
-        let sleep_time = Duration::from_secs(1);
+        let number_of_workers = config.oracle.handle_price_account_update_worker_poll_size;
+        let (sender, receiver) =
+            tokio::sync::mpsc::channel(config.oracle.handle_price_account_update_channel_size);
+        let min_elapsed_time = config.oracle.subscriber_finished_min_time;
+        let sleep_time = config.oracle.subscriber_finished_sleep_time;
 
         handles.push(tokio::spawn(async move {
             loop {
@@ -86,7 +86,7 @@ where
                 .await
                 {
                     tracing::error!(?err, "Subscriber exited unexpectedly");
-                    if current_time.elapsed() < max_elapsed_time {
+                    if current_time.elapsed() < min_elapsed_time {
                         tracing::warn!(?sleep_time, "Subscriber restarting too quickly. Sleeping");
                         tokio::time::sleep(sleep_time).await;
                     }
--- a/src/agent/services/oracle.rs
+++ b/src/agent/services/oracle.rs
@@ -13,7 +13,10 @@ use {
         },
         state::oracle::Oracle,
     },
-    anyhow::{Context, Result},
+    anyhow::{
+        Context,
+        Result,
+    },
     solana_account_decoder::UiAccountEncoding,
     solana_client::{
         nonblocking::{
--- a/src/agent/services/oracle.rs
+++ b/src/agent/services/oracle.rs
@@ -13,10 +13,7 @@ use {
         },
         state::oracle::Oracle,
     },
-    anyhow::{
-        Context,
-        Result,
-    },
+    anyhow::Result,
     solana_account_decoder::UiAccountEncoding,
     solana_client::{
         nonblocking::{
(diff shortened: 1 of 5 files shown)


rodoufu:
I have also this other MR for the project, for which I would love to hear your thoughts.

cc @aditya520 @ali-bahjati

src/agent/services/oracle.rs
```diff
@@ -75,17 +84,34 @@ where
...
+                    let mut receiver = receiver.lock().await;
+                    if let Some(task) = receiver.recv().await {
+                        drop(receiver);
+                        if let Err(err) = task.await {
+                            tracing::error!(%err, "error running price update");
+                        }
+                    }
+                }
+            }));
+        }
```
Riateche, line 114: This implementation basically limits the amount of tasks that can be run at the same time. I think a more straightforward way for doing this is to use a [semaphore](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html). You can create a semaphore permit before spawning a task and move it inside the task. When the task is finished, the permit will automatically be released, allowing more tasks to be spawned. It's simpler because it doesn't require an additional channel and doesn't require using join handles.

src/agent/services/oracle.rs
```diff
@@ -129,14 +156,17 @@ where
...
-                        Oracle::handle_price_account_update(&*state, network, &pubkey, &account)
-                            .await
-                    {
-                        tracing::error!(err = ?err, "Failed to handle account update.");
-                    }
-                });
+                sender
+                    .send(tokio::spawn(async move {
+                        if let Err(err) =
+                            Oracle::handle_price_account_update(&*state, network, &pubkey, &account)
```
Riateche, line 162: I wonder why it's possible to accumulate a lot of long-lived tasks. Maybe we should add a timeout to `Oracle::handle_price_account_update`? Especially considering that now if tasks become stuck, handling price updates can halt completely.

src/agent/services/oracle.rs
```diff
@@ -129,14 +156,17 @@ where
...
+                                .await
+                        {
+                            tracing::error!(?err, "Failed to handle account update");
+                        }
+                    }))
+                    .await
+                    .context("sending handle_price_account_update task to worker")?;
```
Riateche, line 169: I presume we were using `spawn` here to make sure `subscriber()` keeps running while we handling the updates. With this change, `subscriber()` will potentially wait at this `.await` point for previous tasks to complete. This can delay handling new updates. It can also interfere with the pubsub client because we'll not be fetching new updates while we're waiting for old tasks to complete. (It's probably still better than spawning an unlimited amount of tasks.)

rodoufu:
@Riateche I've removed the synchronization and kept only the change to drop the locks before the `.await` that made a great difference here for me.
Could you please take another look?

ali-bahjati (approved):

src/agent/services/oracle.rs
rodoufu: @Riateche I believe we can resolve this thread since I've removed this code, what do you think?

src/agent/services/oracle.rs
rodoufu: @Riateche same here

src/agent/services/oracle.rs
rodoufu: @Riateche and here

ali-bahjati (approved):

PR #154, opened 2025-04-10
chore: Bump Rust version to 1.86.0

merolish:
Trying to use newer versions of crates was choking on the Rust version so I wanted to try moving this forward. Unfortunately there are some solana version mismatches until we can move those forward to version 2.


Cargo.toml
```diff
@@ -27,6 +27,7 @@ pyth-sdk = "0.8.0"
 pyth-sdk-solana = "0.10.4"
 solana-account-decoder = "1.18.8"
 solana-client = "1.18.8"
+solana-pubkey = "2.3.0"
```
ali-bahjati, line 30: hmmm it's strange that this was needed. can you explain a bit?

merolish: So pyth-sdk-solana uses the solana v2 crates but agent itself still uses v1. As best I can tell a newer Rust version changed how imports are handled. The ideal solution is to bump forward the solana sdk totally but for now I wanted to make minimum changes.

ali-bahjati (approved):
@merolish, let's bump the solana later then.

PR #155, opened 2025-04-10
feat(lazer_exporter): Add lazer exporter

merolish:
Add an exporter that periodically polls local state and publishes updates to Lazer relayers. Note that this depends on the symbols endpoint including the Hermes ID so we can map to Lazer ID.


src/agent/services/lazer_exporter.rs
```diff
@@ -0,0 +1,212 @@
...
+                        Some((relayer_url, Ok(msg))) => {
+                            tracing::debug!("Received message from relayer at {relayer_url}: {msg:?}");
+                        }
+                        Some((relayer_url, Err(e))) => {
+                            tracing::error!("Error receiving message from at relayer {relayer_url}: {e:?}");
+                        }
+                        None => {
+                            tracing::error!("relayer connection closed");
+                            return;
+                        }
```
keyvankhademi, line 206: does the thread halt when we get an error?

merolish: Yeah, was wondering if that should crash the agent altogether or have retries in-process.

keyvankhademi: cc: @ali-bahjati

src/agent/services/lazer_exporter.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub description: String,
+    pub asset_type: String,
+    pub exponent: i32,
+    pub cmc_id: Option<u32>,
+    pub interval: Option<String>,
+    pub min_publishers: u16,
+    pub min_channel: String,
+    pub state: String,
+    pub hermes_id: Option<String>,
+}
```
ali-bahjati, line 106: maybe a leave a todo to move it to Lazer's protocol SDK?

src/agent/services/lazer_exporter.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub symbol: String,
+    pub description: String,
+    pub asset_type: String,
+    pub exponent: i32,
+    pub cmc_id: Option<u32>,
+    pub interval: Option<String>,
+    pub min_publishers: u16,
+    pub min_channel: String,
+    pub state: String,
+    pub hermes_id: Option<String>,
```
ali-bahjati, line 105: Is it already in Lazer?

src/agent/services/lazer_exporter.rs
ali-bahjati: the types inside are also not great. we have String for many things but they are more explicit.

src/agent/services/lazer_exporter.rs
```diff
@@ -0,0 +1,227 @@
...
+            Err(e) => {
+                tracing::error!("Failed to fetch Lazer symbols: {e:?}");
+                return;
+            }
+        };
+
+        // Establish relayer connections
+        // Relayer will drop the connection if no data received in 5s
+        let (mut relayer_sender, relayer_receivers) = connect_to_relayers(&config)
+            .await.expect("failed to connect to relayers");
```
ali-bahjati, line 181: hmmm we shouldn't crash if it fails.

src/agent/services/lazer_exporter.rs
```diff
@@ -0,0 +1,227 @@
...
+use url::Url;
+use crate::agent::state;
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct Config {
+    pub history_url: Url,
+    pub relayer_urls: Vec<Url>,
+    pub authorization_token: String,
+    #[serde(with = "humantime_serde")]
+    pub publish_interval_duration: Duration,
```
ali-bahjati, line 30: shall we set a default here? like 10ms?

src/agent/services/lazer_exporter.rs
```diff
@@ -0,0 +1,212 @@
...
+
+        loop {
+            tokio::select! {
+                _ = publish_interval.tick() => {
+                    for (identifier, price_info) in state.get_all_price_infos().await {
+                        if let Some(symbol) = lazer_symbols.get(&identifier.to_string()) {
+                            if let Err(e) = relayer_sender.send_price_update(&PriceFeedDataV1 {
+                                price: Some(Price(NonZeroI64::try_from(price_info.price).unwrap())),
+                                best_ask_price: None,
+                                best_bid_price: None,
```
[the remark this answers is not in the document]
ali-bahjati: I don't know. Typically conf (although we ask for it) is wider than bid-ask by publishers and it makes me a bit worried. let me ask internally.

src/agent/services/lazer_exporter.rs
```diff
@@ -0,0 +1,227 @@
...
+        for (i, receiver) in relayer_receivers.into_iter().enumerate() {
+            stream_map.insert(config.relayer_urls[i].clone(), receiver);
+        }
+
+        let mut publish_interval = tokio::time::interval(config.publish_interval_duration);
+
+        loop {
+            tokio::select! {
+                _ = publish_interval.tick() => {
+                    for (identifier, price_info) in state.get_all_price_infos().await {
```
ali-bahjati, line 192: since we want to run this in high frequency i want to make sure this doesn't become expensive as it might copy the data.

src/agent/services/lazer_exporter.rs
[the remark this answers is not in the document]
ali-bahjati: i still prefer None here just because bid ask might not be symmetric. we can do it on a newer update protocol.

src/agent/services/lazer_exporter.rs
```diff
@@ -0,0 +1,227 @@
...
+async fn fetch_symbols(history_url: &Url) -> Result<Vec<SymbolResponse>> {
+    let mut url = history_url.clone();
+    url.set_scheme("http").unwrap();
+    url.set_path("/history/v1/symbols");
+    let client = Client::new();
+    let response = client
+        .get(url)
+        .send()
+        .await?
+        .text()
```
ali-bahjati, line 117: you should be able to json directly here instead of converting later.

ali-bahjati (commented):
Nice! I'm happy that you left some works for future (retry/reliability and continuous state polling). I'll approve if things apart from that get resolved.

ali-bahjati:
ah another comment, please fix CI (formatting/..) and clippy warnings.

ali-bahjati:
@keyvankhademi yeah sure thing. It's still better to make changes in multiple PRs and merge this one. we can merge the code and mark it as alpha (@merolish let's do it). no one can publish data until tokens are given out.

src/agent/services/lazer_exporter.rs
merolish: It is not, but I can put up a PR to add it in history-service.

src/agent/services/lazer_exporter.rs
merolish: I copied from https://github.com/pyth-network/pyth-lazer/blob/main/history-service/src/api.rs#L91 ; sorry, not clear what they should be.

src/agent/services/lazer_exporter.rs
merolish: Added a brief PR to add it.

src/agent/services/lazer_exporter.rs
merolish: Pushing fix.

src/agent/services/lazer_exporter.rs
merolish: Yep, pushing fix.

src/agent/services/lazer_exporter.rs
merolish: Fixed!

ali-bahjati (approved):

src/agent/services/lazer_exporter.rs
```diff
@@ -0,0 +1,281 @@
...
+        let mut publish_interval = tokio::time::interval(config.publish_interval_duration);
+
+        loop {
+            tokio::select! {
+                _ = publish_interval.tick() => {
+                    // TODO: This read locks and clones local::Store::prices, which may not meet performance needs.
+                    for (identifier, price_info) in state.get_all_price_infos().await {
+                        if let Some(symbol) = lazer_symbols.get(&identifier.to_string()) {
+                            if let Err(e) = relayer_sender.send_price_update(&PriceFeedDataV1 {
+                                price: Some(Price(NonZeroI64::try_from(price_info.price).unwrap())),
```
ali-bahjati, line 249: what if the price is zero? are we sure it won't happen?

merolish: Excellent question, no idea, and I was curious on this enforcement in the Lazer messages. In the general case there are instruments out there (e.g. futures spreads) that can have zero or negative prices, but in terms of anything that anyone will reasonably send to agent, perhaps just warning log and drop a zero price message?

ali-bahjati: I think we support negative but just not zero. This came from Pythnet.

What i like you to do is at least not crash the whole process when it happens and gracefully handle it instead.

ali-bahjati:
@merolish Btw, looking at this PR again. Can you update the config.toml reference in your future PRs.

merolish:
Oops, sorry about that.

PR #158, opened 2025-04-23
feat: publish signed Lazer transactions

merolish:
Initial attempt at publishing a SignedLazerTransaction on each publish interval..

### DIFF
--- a/src/agent/services/lazer_exporter.rs
+++ b/src/agent/services/lazer_exporter.rs
@@ -47,9 +47,17 @@ pub struct Config {
     pub relayer_urls:              Vec<Url>,
     pub publisher_id:              u32,
     pub authorization_token:       String,
-    pub publisher_signing_key:     Vec<u8>,
+    publisher_keypair:             PublisherKeypair,
     #[serde(with = "humantime_serde", default = "default_publish_interval")]
     pub publish_interval_duration: Duration,
+}
+
+#[derive(Clone, Deserialize)]
+struct PublisherKeypair(Vec<u8>);
+impl std::fmt::Debug for PublisherKeypair {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "PublisherKeypair(redacted)")
+    }
 }
 
 fn default_publish_interval() -> Duration {
@@ -244,13 +252,13 @@ mod lazer_exporter {
 
         // Establish relayer connections
         // Relayer will drop the connection if no data received in 5s
-        let (mut relayer_sender, relayer_receivers) = connect_to_relayers(&config).await?;
+        let (mut relayer_sender, relayer_receivers) = connect_to_relayers(config).await?;
         let mut stream_map = StreamMap::new();
         for (i, receiver) in relayer_receivers.into_iter().enumerate() {
             stream_map.insert(config.relayer_urls[i].clone(), receiver);
         }
 
-        let keypair = Keypair::from_bytes(&config.publisher_signing_key)?;
+        let keypair = Keypair::from_bytes(&config.publisher_keypair.0)?;
         let mut publish_interval = tokio::time::interval(config.publish_interval_duration);
 
         loop {
--- a/src/agent/services/lazer_exporter.rs
+++ b/src/agent/services/lazer_exporter.rs
@@ -13,7 +13,8 @@ use {
         SinkExt,
     },
     http::HeaderValue,
-    pyth_lazer_protocol::publisher::PriceFeedDataV1,
+    protobuf::Message as ProtobufMessage,
+    pyth_lazer_publisher_sdk::transaction::SignedLazerTransaction,
     reqwest::Client,
     serde::Deserialize,
     std::{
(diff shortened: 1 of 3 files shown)


src/agent/services/lazer_exporter.rs
```diff
@@ -241,32 +251,73 @@ mod lazer_exporter {
...
 
         loop {
             tokio::select! {
                 _ = publish_interval.tick() => {
+                    let now = Utc::now().timestamp_micros();
+                    let publisher_timestamp = MessageField::some(Timestamp {
+                        seconds: now / 1_000_000,
+                        nanos: (now % 1_000_000 * 1000) as i32,
+                        special_fields: Default::default(),
+                    });
```
keyvankhademi, line 265: I've seen this multiple times in different locations. Maybe we add it to the sdk (if it's not there yet) and re use it.
@darunrs @Riateche

keyvankhademi:
lgtm!

src/agent/services/lazer_exporter.rs
merolish: Yeah I didn't see any adapters/helpers in rust-protobuf so just a matter of hooking up our own.

darunrs: You can do [Timestamp::now()](https://docs.rs/protobuf/latest/protobuf/well_known_types/timestamp/struct.Timestamp.html) and it will automatically make one with the current time. MessageField::some(Timestamp::now()) is not wordy at that point.

src/agent/services/lazer_exporter.rs
```diff
@@ -48,7 +45,9 @@ use {
 pub struct Config {
     pub history_url:               Url,
     pub relayer_urls:              Vec<Url>,
+    pub publisher_id:              u32,
     pub authorization_token:       String,
+    pub publisher_signing_key:     Vec<u8>,
```
darunrs, line 50: Make this a new type or struct and don't derive Debug on it. We should then get errors if we ever try to print this. Add ignore debug here to make sure it's never printed.

src/agent/services/lazer_exporter.rs
```diff
@@ -241,32 +251,73 @@ mod lazer_exporter {
...
-                                best_ask_price: None,
-                                best_bid_price: None,
-                                price_feed_id: PriceFeedId(symbol.pyth_lazer_id),
-                                publisher_timestamp_us: TimestampUs::now(),
-                                source_timestamp_us: TimestampUs(price_info.timestamp.and_utc().timestamp_micros() as u64),
-                            }).await {
-                                tracing::error!("Error sending price update to relayer: {e:?}");
-                                bail!("Failed to send price update to relayer: {e:?}");
-                            }
+                            let source_timestamp_micros = price_info.timestamp.and_utc().timestamp_micros();
```
darunrs, line 277: What type is this? There is an impl From for SystemTime. Otherwise, let's add a From for DateTime. Also, instead of UTC, convert to epoch instead. I am pretty sure thats still UTC but we should be explicit.

src/agent/services/lazer_exporter.rs
merolish: Aaaaaahhh it's in well_known_types_util/timestamp.rs instead of well_known_types/timestamp.rs where the struct is defined. I hate hate hate this about Rust, there's no actual class interface in the code and so you end up having to dig through docs.rs which I don't find terribly ergonomic.

darunrs (approved):
Nice start! There's a couple things that Pyth Agent will need to do:

1. Receive Publisher Updates from a WS
2. Collect, convert, and dedupe updates
3. Batch updates into a payload on an interval
4. Encode payload, sign payload, make transaction with encoded payload and signature 
5. Encode transaction and send it

Hopefully that helps you plan out abstractions and how to break code up!

src/agent/services/lazer_exporter.rs
darunrs: Hm? It is `well_known_types/timestamp.rs`?
https://github.com/pyth-network/pyth-lazer/blob/main/internal-protocol/src/lib.rs#L1

darunrs: Also I think `new()` also works.

merolish: I just meant that in rust-protobuf the Timestamp::now() method is defined in a separate util module, wouldn't have thought to look for it elsewhere. And thanks for the helper, although I don't think I need to touch TimestampUs at all in agent.

merolish:
In the agent 1 is done with an rpc interface and 2 is done by writing to a "local state" cache, so I'm not touching either at the moment. For 3 the "exporter" takes this latest cache of prices and batch publishes, so I'm following this pattern for now, just sending to the Lazer relayers instead of pythnet. We're just handling 3/4/5 here.

src/agent/services/lazer_exporter.rs
merolish: This is a chrono::NaiveDateTime that is used in the existing local cache struct, so I didn't want to change it for now. If we extricate the pythnet part or rewrite from scratch then yeah, makes sense.

PR #159, opened 2025-04-25
chore: Bump Rust edition and Solana crates

merolish:
- Bump Rust edition to 2024
- Bump Solana client crates to 2.2 to align with pyth-sdk-solana
- fmt (mainly rearrangement of imports)
- Some clippy allowances
- ed25519-dalek v2 updates

### DIFF
--- a/src/agent/services/exporter.rs
+++ b/src/agent/services/exporter.rs
@@ -223,9 +223,9 @@ mod exporter {
                 },
             },
             state::exporter::{
+                Exporter,
                 get_publish_keypair,
                 publish_batches,
-                Exporter,
             },
             utils::rpc_multi_client::RpcMultiClient,
         },
--- a/src/agent/services/exporter.rs
+++ b/src/agent/services/exporter.rs
@@ -210,6 +210,7 @@ where
     handles
 }
 
+#[allow(clippy::module_inception)]
 mod exporter {
     use {
         super::NetworkState,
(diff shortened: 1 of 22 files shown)


keyvankhademi (approved):

keyvankhademi:
lgtm. But wait for someone else to review too.

src/agent/services/exporter.rs
```diff
@@ -210,6 +210,7 @@ where
     handles
 }
 
+#[allow(clippy::module_inception)]
```
darunrs, line 213: Why are we getting this clippy problem now?

darunrs (approved):
I'm a bit nervous this doesn't run any tests as part of CI. Does Agent not have any tests at all?

darunrs:
Also this probably qualifies for a version bump too.

src/agent/services/exporter.rs
merolish: As far as I can tell these have been existing issues and clippy isn't an existing PR check.

merolish:
Good point on the version bump. Unit and integration tests are run; see workflows/rust.yaml, although I cannot vouch for how thorough they are and they also don't include the lazer side yet. We will need to properly test this prior to distributing the new version.

merolish:
(The integration test stuff is in integration-test, which mocks a pythnet node in Python.)

PR #166, opened 2025-06-16
Lazer exporter fixes

merolish:
- Use Lazer pubkey as authorization token for relayer
- Fetch symbol list from history service periodically

### DIFF
--- a/src/agent/services/lazer_exporter.rs
+++ b/src/agent/services/lazer_exporter.rs
@@ -1,6 +1,7 @@
 use {
     crate::agent::state,
     anyhow::{
+        Context,
         Result,
         anyhow,
         bail,
@@ -9,6 +10,11 @@ use {
         ExponentialBackoffBuilder,
         backoff::Backoff,
     },
+    base64::{
+        Engine,
+        prelude::BASE64_STANDARD,
+    },
+    ed25519_dalek::SigningKey,
     futures_util::{
         SinkExt,
         stream::{
@@ -25,6 +31,7 @@ use {
         Deserialize,
         Serialize,
     },
+    solana_sdk::signature::keypair,
     std::{
         path::PathBuf,
         sync::Arc,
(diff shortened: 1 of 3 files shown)


src/agent/services/lazer_exporter.rs
```diff
@@ -248,17 +260,44 @@ async fn fetch_symbols(history_url: &Url) -> Result<Vec<SymbolResponse>> {
...
+    };
+
     // can safely drop first receiver for ease of iteration
     let (relayer_sender, _) = broadcast::channel(RELAYER_CHANNEL_CAPACITY);
 
     for url in config.relayer_urls.iter() {
         let mut task = RelayerSessionTask {
             url:      url.clone(),
-            token:    config.authorization_token.to_owned(),
+            token:    BASE64_STANDARD.encode(signing_key.verifying_key().to_bytes()),
```
darunrs, line 300: Can we separate the signing key from the verifying key really early in the process, and pass them around separately? This is like one typo away from accidentally encoding and sending the private key, which makes me nervous. In fact, I wonder if we can put some guards in to prevent the private key from being encoded after its read or something.

darunrs: Also why are we setting the public key as the token? The plan was [to retain access tokens unitl we can remove them](https://linear.app/douro-labs/issue/PF-136/publisher-access-token-deprecation).

merolish: The latter was an ask from Ali.

keyvankhademi: @darunrs we want pythnet publishers to start publishing asap. We will remove this when we implement the changes in the relayer. For now we still need to use tokens to match with publishers.

darunrs: Gotcha. Thanks for filling me in. That makes sense, and I don't think it is problematic to make them shared in the short term.

darunrs: Regarding private key leaking, we can override (or not derive) things like Debug, Serialize, Encode with something that produces dummy values for the signing key by wrapping it in a struct or type and making it a private member of the struct.

keyvankhademi (commented):
lgtm but let's have @ali-behjati approve it.

